PR #115, opened 2024-09-05
Add Token Caching

spjmurray:
Transpires that validating tokens is very expensive, so add a cache to identity to speed this up.

### DIFF
--- a/charts/identity/Chart.yaml
+++ b/charts/identity/Chart.yaml
@@ -4,8 +4,8 @@ description: A Helm chart for deploying Unikorn's IdP
 
 type: application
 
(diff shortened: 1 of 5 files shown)


spjmurray:
Before:
Screenshot from 2024-09-05 15-58-55

After:
Screenshot from 2024-09-05 15-59-10

spjmurray:
With client side caching of the userinfo data... PREPARE SHIP FOR LUDICROUS SPEED!
Screenshot from 2024-09-05 16-23-48


That's 464ms to 6ms... take that in, that's 1.2% of the time, or 77x faster.

PR #125, opened 2024-10-15
Create RBAC for Compute Clusters

nsricardor:
This pull request introduces RBAC for the following endpoints: computeclusters, servers, and securitygroups to ensure users have appropriate access levels based on their roles.

### DIFF
--- a/charts/identity/values.yaml
+++ b/charts/identity/values.yaml
@@ -147,8 +147,6 @@ roles:
         applications: [read]
         applicationsets: [create,read,update,delete]
         computeclusters: [create,read,update,delete]
-        servers: [create,read,update,delete]
-        securitygroups: [create,read,update,delete]
   # A reader can view projects they are a member of and view
   # kubernetes clusters.
   reader:
@@ -163,8 +161,6 @@ roles:
         applications: [read]
         applicationsets: [read]
         computeclusters: [read]
-        servers: [read]
-        securitygroups: [read]
 
 ingress:
   # Sets the ingress class to use.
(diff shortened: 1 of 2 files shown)


charts/identity/values.yaml
```diff
@@ -140,6 +146,9 @@ roles:
         kubernetesclusters: [create,read,update,delete]
         applications: [read]
         applicationsets: [create,read,update,delete]
+        computeclusters: [create,read,update,delete]
+        servers: [create,read,update,delete]
```
spjmurray, line 150: End users shouldn't be able to create and delete random servers, they should do it via a managed service.

charts/identity/values.yaml
```diff
@@ -153,6 +162,9 @@ roles:
         kubernetesclusters: [read]
         applications: [read]
         applicationsets: [read]
+        computeclusters: [read]
+        servers: [read]
```
spjmurray, line 166: And delete these two too!

spjmurray (changes requested):
Just a couple tweaks to tighten up security.

While I'm here, and it's fresh in my mind... When we create Kubernetes clusters, the server component also creates `identitiy` and possibly `physicalnetwork` resources as the user (due to the token being available - legacy behaviour from "the before times"). 

This too is a security hole that needs to be closed up and the server component itself, or perhaps the controller can handle.  I'll do this sometime soon. This will obviously affect compute too as it'll need all that to be configured before it can proceed with security groups and servers.  Something to be aware of!

spjmurray (approved):

spjmurray:
Ah shit, @nsricardor you need your commits to be signed to prevent impersonation...

PR #131, opened 2024-11-20
Feat/onboarding

AdamFlanagan-Nscale:


README.md
```diff
@@ -30,9 +30,9 @@ The identity service provides some generic providers, Google, and Microsoft, whi
 
 You can _bring your own_ by providing:
 
-* And OIDC compliant issuer endpoint
-* A client ID
-* A client secret
+- And OIDC compliant issuer endpoint
```
spjmurray, line 33: Don't be reformatting the documentation please (yeah, yeah, blame your editor), it's unnecessary change that makes my life harder and it makes this file "special" with regard to all other documentation.

spjmurray: That said, there is a typo here if you read it 😄

charts/identity/templates/identity/deployment.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         - --host=https://{{ include "unikorn.identity.host" . }}
         - --jose-tls-secret=unikorn-identity-jose-tls
         - --refresh-token-duration={{ printf "%dh" (mul .Values.issuer.maxTokenDurationDays 24) }}
+        - --allow-new-user-organizations={{ .Values.identity.allowNewUserOrganizations | default false }}
```
spjmurray, line 27: The `default` is superfluous here and just increases the logic when `values.yaml` already handles this.

README.md
```diff
@@ -287,6 +287,61 @@ kubectl unikorn create group \
...
+
+```bash
+kubectl create -n unikorn-identity -f - <<EOF
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: unikorn-ui
+  namespace: unikorn-identity
+spec:
+  commonName: unikorn-ui
```
spjmurray, line 313: I think using `unikorn-ui` here is misleading as it logically leads people to the actual UI, which doesn't use this!

pkg/openapi/server.spec.yaml
```diff
@@ -60,6 +60,25 @@ paths:
           $ref: 'https://raw.githubusercontent.com/unikorn-cloud/core/main/pkg/openapi/common.spec.yaml#/components/responses/unauthorizedResponse'
         '500':
           $ref: 'https://raw.githubusercontent.com/unikorn-cloud/core/main/pkg/openapi/common.spec.yaml#/components/responses/internalServerErrorResponse'
+  /api/v2/create-account:
```
spjmurray, line 63: `v1` please, adding things is okay, removing or changing them is bad.

pkg/handler/organizations/client.go
```diff
@@ -245,3 +245,8 @@ func (c *Client) Update(ctx context.Context, organizationID string, request *ope
 
 	return nil
 }
+
+// Special case for onboarding.
+func Convert(in *unikornv1.Organization) *openapi.OrganizationRead {
```
spjmurray, line 250: You're defining the same thing twice here, just capitalize the existing one!

pkg/oauth2/oauth2.go
```diff
@@ -825,8 +829,8 @@ func (a *Authenticator) OIDCCallback(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !userExists {
-		authorizationError(w, r, state.ClientRedirectURI, ErrorAccessDenied, "user not found")
+	if !userExists && !a.options.AllowNewUserOrganizations {
```
spjmurray, line 832: I think you may want to not do the RBAC check at all here rather than wasting compute cycles!

charts/identity/templates/identity/deployment.yaml
spjmurray: The name is a bit off here... hmmm... `--authenticate-unknown-users` perhaps is more descriptive?

spjmurray (changes requested):
Looks good so far, just a few naming and cleanliness requests and then I'll get down and dirty with the actual algorithms...

pkg/middleware/openapi/openapi.go
```diff
@@ -201,14 +223,17 @@ func (v *Validator) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx = authorization.NewContextWithUserinfo(ctx, v.userinfo)
 
 	if v.userinfo != nil {
-		// The organizationID parameter is standardized across all services.
-		acl, err := v.authorizer.GetACL(ctx, params["organizationID"], v.userinfo.Sub)
```
spjmurray, line 205: Okay, so because of how Go maps work, this will return the "zero value" if the key does not exists i..e an empty string...

So I guess the `rbac.GetACL`  call can just return nowt if this condition holds, which is as KISS as you can get.

pkg/handler/onboarding/client.go
```diff
@@ -0,0 +1,169 @@
...
+func (c *Client) validateAndGetAdminRole(ctx context.Context) (*unikornv1.Role, error) {
+	var roleList unikornv1.RoleList
+	if err := c.client.List(ctx, &roleList, &client.ListOptions{
+		Namespace: c.namespace,
+	}); err != nil {
+		return nil, errors.OAuth2ServerError("failed to list roles").WithError(err)
+	}
+
+	for _, role := range roleList.Items {
+		if role.Labels[constants.NameLabel] == "administrator" {
```
spjmurray, line 129: Nerp, hard coding, add a flag to identify the default role.

pkg/util/wait/resource.go
```diff
@@ -0,0 +1,112 @@
+/*
```
spjmurray, line 1: Good, but it's going off on a tangent as regards other things... so I'm thinking something like:

```go
import (
  "context"

  "github.com/unikorn-cloud/core/pkg/util/retry"

  corev1 "k8s.io/api/core/v1"

  "sigs.k8s.io/controller-runtime/pkg/client"
)

type Validator interface {
  Valid(resource unikornv1core.ManagableResourceInterface) bool
}

type AvailableCondition struct {}

func NewAvailableConditionValidator() *AvailableCondition {
  return &AvailableCondition{}
}

func (c *AvailableCondition) Valid(resource unikornv1core.ManagableResourceInterface) error {
  condition, err := resource.StatusConditionRead(unikornv1core.ConditionAvailable)
  if err != nil {
    return err
  }

  if condition.Status != corev1.ConditionTrue {
    return fmt.Errorf("%w: resource not avialable", ErrCondition)
  }

  return nil
}

func WaitForResource(ctx context.Context, cli client.Client, key *client.ObjectKey, resource unikornv1core.ManagableResourceInterface, validators ...Validator) error {
  callback := func() error {
    if err := cli.Get(ctx, key, resource); err != nil {
      return err
    }

    for _, validator := range validators {
      if err := validator.Valid(resource); err != nil {
        return err
      }
  }

  return nil
}
```

Or something to that effect, then you can go....

```go
myCtx, cancel := context.WithTimeout(ctx, time.Minute)
defer cancel()

var resource unikornv1.Organization

if err := WaitForResource(myCtx, cli, client.ObjectKey{Namespace: foo, Name: bar}, &resource, NewAvailableConditionValidator()); err != nil {
  // don't clean it up, we want to examine it to work out why and fix the bug!
  return err
}
```

spjmurray (changes requested):
Getting there!

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -220,6 +220,19 @@ func (a *Authorizer) GetACL(ctx context.Context, organizationID, subject string)
 		return nil, errors.OAuth2ServerError("failed to create identity client").WithError(err)
 	}
 
+	if organizationID == "" {
```
spjmurray, line 223: I think the key thing you are actually trying to solve is that "client X doesn't need an ACL".  At which point just adding in some logic that reads from a new and totally unused API is a little redundant.  Why not just add an option to the authorizer that allows you to disable the ACL read?  See a little sleep would have saved you a load of work 😸

spjmurray: Or does your client need an ACL?  I'm utterly blind without seeing what you are actually doing.

README.md
```diff
@@ -287,6 +287,62 @@ kubectl unikorn create group \
 > Individual services will document their CN and role requirements.
 > All official Unikorn Cloud services will have their roles pre-defined by this repository.
 
+### Enabling User Sign-ups
```
spjmurray, line 290: I wouldn't bother documenting this at all.   Delete, delete, delete.

charts/identity/values.yaml
```diff
@@ -121,6 +125,14 @@ roles:
     scopes:
       global:
         kubernetes:clusters: [read]
+  # A create account service is a role that can create organizations.
```
spjmurray, line 128: Can this not live in your on-boarding helm chart?  It doesn't feel "official".  Just create a template that makes a Role in the `unikorn-identity` namespace.

spjmurray (changes requested):

spjmurray (approved):

PR #244, opened 2025-03-26
Remove overly generic go build rule

squaremo:
The pattern `bin/%` matches e.g.,
`bin/amd64-linux/unikorn-controller`, but will bind `$*` to `amd64-linux/unikorn-controller` in the recipe and thereby fail.

The (GNU) make pre-installed in MacOS falls foul of this, but brew-installed GNU make runs OK. Still, let's avoid ambiguity if we don't need it.


spjmurray (approved):
This is fine... now do every other repo 😜

PR #254, opened 2025-05-07
Add Token Hashing

### LINKED ISSUE #253: Hash Tokens in User Records
While keeping active tokens in the `User` record is great for a quick and easy way to compare what is handed over at the API, it's also a honey pot of information.  Refesh tokens are bound to the client ID, and issuing a token needs the client secret, which is another problem in itself!  So really the only thing to worry about here is access tokens that are alive for an hour, or service account tokens which is slightly more of an issue being live for up to 3 months.

So basically run these through SHA2.  For a transparent upgrade, we can probably prefix the entry with `SHA2-512:<hash>` to identify the algorithm, defaulting to plaintext for backwards compatibility.  We can even extend this to `SHA2-512:512:<hash>` to identify the number of rounds to subject the hash to for improved security, and also allow us to increase the number of rounds in the future to thwart crypanalysis.

spjmurray:
Reduces the attack surface by obfuscating tokens using SHA256 and PBKDF2, thus they are only ever persisted client side, we continue to maintain a plaintext in-memory cache for performace reasons.

Fixes #253


spjmurray:
Let's see what Pryor is like as a reviewer... 😸

PR #260, opened 2025-05-15
Reduce System Account Coupling

spjmurray:
We define a single system account role that is used by (at present) the Kubernetes and Compute services.  This has a few drawbacks: first you have to modify identity to enable new functionality in either of those services, and secondly you may inadvertently give away permissions that aren't strictly necessary.

What this does is allows each service to define a SystemAccount resource in their own Helm chart, linked to their own role.  It also means 3rd party services can define all their own stuff in situ without having to tell the Identity service about itself, or know where the Identity service lives in order to install custom roles.

Security wise, anyone who can create a certificate, role and system account can then issue a token to do what they want, so I guess limiting who can do those operations brings security, or perhaps having a PSK to validate the resource is trusted via the PSK.  Prior to this a user would have to issue a certificate, create a role (or just reference the uber user one - perhaps we should grant this implicitly via a magical \*.\* ACL) and modify the identity deployment.  So perhaps backward a little bit, but it does offer operational and development benefits.


spjmurray:
See what you make of this... sister pulls exist in https://github.com/nscaledev/uni-kubernetes/pull/244 and https://github.com/nscaledev/uni-compute/pull/62

PR #268, opened 2025-06-05
Add Tweakable Resource Limits

spjmurray:
Allow all services to have their resource allocations modified.  This mostly keeps things as they were, but only specifies limits not for guaranteed QoS scheduling.  The identity server has been expanded by 10x as memory utilization does tend to increase quickly when placed on the public internet!  The long term solution to growing numbers of resources and cache sizes is probably an external database, however that breaks the whole cascading cleanup thing when using namespaces...


spjmurray:
Tested with `deploy -t | kubectl -n unikorn-identity diff -f -` and does only what it should.

PR #278, opened 2025-07-01
Principal Propagation

spjmurray:
This does a few things to enable principal propagation throughout the system.

Defines what a principal is, organization, project and end-user identity tuple and provides some machinery to stash and retrieve from the context.

Next it adds in some openapi middleware to extract the principal header from a HTTP request, and validate it against the (validated by mTLS) client certificate and stash it in the context.

We provide interfaces for clients to use.  For handlers we move the identity bits of ObjectMeta out of core into here, then have it implicitly fill in identity information for created Kubernetes resources, and also provide an interface to get access to the principal (from the context), for use with things like quota allocations.  The final interface is used by controllers who consume said resources and proxy requests on to other services, this generically extracts the principal from a resources, encodes it and injects it into a remote HTTP request as a header.


pkg/client/base.go
```diff
@@ -0,0 +1,142 @@
+/*
```
spjmurray, line 1: The point of this is to reduce duplication in identity/region/compute/kubernetes, same stuff, just returns a different client type

pkg/middleware/openapi/openapi.go
```diff
@@ -117,10 +131,37 @@ func (w *bufferingResponseWriter) StatusCode() int {
 }
 
 func (v *Validator) validateRequest(r *http.Request, route *routers.Route, params map[string]string) (*openapi3filter.ResponseValidationInput, error) {
+	// This authorization callback is fired if the API endpoint is marked as
```
spjmurray, line 134: The changes here are mostly to follow the model defined in https://github.com/nscaledev/uni-specifications/blob/main/specifications/identity/authentication_modes.md.  The comments should be enough to explain some of the decisions....

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -118,6 +118,44 @@ func (t *requestMutatingTransport) RoundTrip(req *http.Request) (*http.Response,
...
+
+	client, err := identity.HTTPClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Whe need to mutate the request to do trace context propagation and
+	// client certificate propagation if it's a token bound to an X.509
+	// certificate.
+	mutator := func(req *http.Request) error {
```
spjmurray, line 135: This is primarily to remove an "if has bearer token" in the client.  Now it's required everywhere except this special case where the OIDC code provides it for you.

pkg/middleware/openapi/openapi.go
spjmurray: See also https://github.com/nscaledev/uni-specifications/blob/main/specifications/identity/principals.md

pkg/client/interfaces.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 package client
 
-// AccessTokener provides an interface to retrieve an access token.
-type AccessTokener interface {
+// AccessTokenGetter provides an interface to retrieve an access token.
```
spjmurray, line 19: Renamed for parity with other services.

squaremo:
I have read this to have some confidence I understand what's going on:

 - BaseClient and Builder[T] adapt between the generic bits of clients and the types local to each service's openapi package. It's a nice way to split that.
 - Lots of rearrangement in the pkg/handler, to create and update the standard object metadata. These all _look_ similar, enough so that I made a quixotic attempt to refactor the changed bits -- but they are all just individual enough to need individual treatment
 - The authorisation middleware (pkg/middleware/openapi) changes to enforce the rule that any request coming from another service must have a principal attached.

On that last one: there's two big changes here, and a fair bit of rearrangement. Change one: a principal is required, and this is checked inside the authorisation callback (effectively https://github.com/nscaledev/uni-identity/pull/278/files#diff-7a10793e5382b9b7a0095ec974143cbd49bbea4055366ad883c3308eadf1cad7R149); change two: because a principal is required on requests, there needs to be some sleight of hand for /userinfo and /acl, since they are called _during_ authorisation.

It is quite difficult to tell if this last change is correct, because the decision of whether a request is authorised is distributed amongst various funcs. I have some suggestions which might help:

 - We have a test suite which stands up the identity service and another service, then makes clients with and without the required properties and makes sure they are/are not authorised. This would demonstrate that the code is correct. It'll also help validate later changes, too.
 - If possible, the decision of what kind of client is making a request, and therefore what must be present in the request, is made early, and each alternative has its own explicit call chain.

pkg/middleware/openapi/openapi.go
```diff
@@ -157,59 +198,119 @@ func (v *Validator) validateResponse(w *bufferingResponseWriter, r *http.Request
...
+
+// validateAndAuthorize performs OpenAPI schema validation of the request, and also
+// triggers an authentication callback when the APi is marked as requiring it.
+// This will read the request body from the original and replace it with a buffer.
+// As we are doing a shallow copy to inject authentication context information you
+// must use the returned request for the HTTP handlers.
+func (v *Validator) validateAndAuthorize(ctx context.Context, r *http.Request, route *routers.Route, params map[string]string) (*http.Request, *openapi3filter.ResponseValidationInput, error) {
+	// If mTLS is in use, then the access token *may* be bound to the X.509 private key,
+	// but only in the case where a service is using a client credentials grant.
+	// As all services act on behalf of clients, we only want the client certificate to
```
squaremo, line 267:
```suggestion
	// As all services act on behalf of clients, we want only the client certificate to
```

Is this change correct, as opposed to 

> As all services act on behalf of clients, we want the client certificate to
> be propagated **only** to the identity service during authentication (userinfo call) and
> authorization (ACL call)

pkg/middleware/openapi/openapi.go
```diff
@@ -157,59 +198,119 @@ func (v *Validator) validateResponse(w *bufferingResponseWriter, r *http.Request
...
 	// If any authentication was requested as part of the route, then update anything
 	// that needs doing.
 	if v.info != nil {
 		// Propagate authentication/authorization info to the handlers
-		// and the ACL layer to use.
+		// for the pursposes of auditing and RBAC.
 		ctx = authorization.NewContext(ctx, v.info)
+		ctx = rbac.NewContext(ctx, v.acl)
+
+		// Trusted clients using mTLS must provide principal information in the headers.
```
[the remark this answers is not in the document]
squaremo: One downside of the sentinel is that it would introduce the possibility of accidentally propagating a valid but incorrect value for the principal to other API calls. So I prefer having the special case for service-to-service calls that doesn't require a principal.

squaremo (approved):
Nothing show-stopping, but please have a scan through my comments. There's a handful of suggested changes; and a (probably harmless, but still ..) duplicate bit of code to remove.

Please no more +1600 -500 in a single commit! (unless it's a mechanical change)

pkg/handler/organizations/client.go
```diff
@@ -271,17 +272,20 @@ func (c *Client) Get(ctx context.Context, organizationID string) (*openapi.Organ
...
 
-	out := &unikornv1.Organization{
-		ObjectMeta: conversion.NewObjectMetadata(&in.Metadata, c.namespace, info.Userinfo.Sub).Get(),
+	if err := common.SetIdentityMetadata(ctx, &out.ObjectMeta); err != nil {
+		return nil, errors.OAuth2ServerError("failed to set identity metadata").WithError(err)
 	}
 
 	out.Spec.Tags = conversion.GenerateTagList(in.Metadata.Tags)
 
+	if err := common.SetIdentityMetadata(ctx, &out.ObjectMeta); err != nil {
```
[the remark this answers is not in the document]
spjmurray: It will still set the creator for auditing purposes.

pkg/util/tls.go
```diff
@@ -33,6 +33,17 @@ var (
 	ErrClientCertificateError = errors.New("client certificate error")
 )
 
+// HasClientCeritifcateHeader checks if mTLS is in use, thus we are being called
+// from a trusted service.
+func HasClientCeritifcateHeader(header http.Header) bool {
```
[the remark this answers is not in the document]
spjmurray: Probably, it should have already been read and rejected if not valid by the point that this is invoked , and then we could just pass the raw certificate directly into the extractPrincipal function.  I think it's clearer this way, rather that going "there's an error, so let's generate one".

spjmurray:
New commit only updates a few comments to make them clearer, nothing technical so, let's go baby!

pkg/middleware/openapi/openapi.go
squaremo: OK; I ask because the request with that context is returned and used in subsequent calls. So if the certificate is only wanted for these calls, that context should be discarded afterward.

PR #282, opened 2025-07-16
Use httpsnoop.Wrap for response capture

squaremo:
Wrapping http.ResponseWrite is fraught (see
https://github.com/felixge/httpsnoop/tree/v1.0.4?tab=readme-ov-file#why-this-package-exists).

This uses httpsnoop.Wrap to wrap the response while trying to capture
the bytes written to it, etc. This overcomes some gaps in the prior
implementation -- which could surely have been papered over, but why
not use the battle-tested library to do it once and for all.

I've supplied a wrapper for ReadFrom because httpsnoop will call it if
it exists on the ResponseWriter, and the implementation in `http` does
implement it. (It was fine _not_ having it on the prior implementation,
because io.Copy would fall back to buffered writes anyway.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.24.2
 
 require (
 	github.com/coreos/go-oidc/v3 v3.14.1
-	github.com/felixge/httpsnoop v1.0.4
 	github.com/getkin/kin-openapi v0.132.0
 	github.com/go-chi/chi/v5 v5.2.2
 	github.com/go-jose/go-jose/v4 v4.1.1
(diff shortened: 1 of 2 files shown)


spjmurray (commented):
Interesting... a thought occurs, there is a second implementation in core/pkg/server/middleware/types.go that may do the same thing, or nearly, so should we remove the duplication and fix once, fix everywhere?

squaremo:
The response writer wrapper in core/pkg/server/middleware/types.go captures the body, but it only uses the length of the buffer -- i.e., the number of bytes written. httpsnoop already has a func which does exactly this, so in https://github.com/nscaledev/uni-core/pull/155 I used that.

As a result, there seems little reason to move the abstraction here, there, when it's used here. So I'll leave this PR as-is.

squaremo:
> Gah, slight snag when I was testing this: the `types.go` you removed in core is a requirement of identity's audit logger, so that will need updating too. That captures the body too in order to extract the metadata ID of resource creation requests, so I guess the body capturing variant should live in core perhaps due to sharing and caring?

Ops, sorry about that. Is it used anywhere else? I can just expand the use of httpsnoop use in identity; but I don't mind moving it here either, even though it'd not be used (I would still use httpsnoop.CaptureMetrics here).

spjmurray:
Up to you, in @squaremo we trust, but like add a `replace github.com/unikorn-cloud/core => ../core` or whatevs to `go.mod` to predict the future.

spjmurray (approved):
Basic Identity tests show nothing out of the ordinary in the logs to do with OpenAPI, take it away!

PR #288, opened 2025-08-20
Fix Project Namespace Lookups

spjmurray:
Other services may create namespaces for their work which may also feature the organization and project ID labels, which will alias project namespaces and cause errors as multiple are returned.  Luckily we already have a kind label to differentiate namespaces types so use this. Additionally make the lookup function shared so it can be consumed by bothe compute and Kubernetes.

Eventually this problem will go away once project namespaces are history.


spjmurray:
Tested locally with an aliasing namespace, it returned an error, as expected given the recreation hypothesis.

Adding in the fix solved the issue.

PR #294, opened 2025-09-03
External Finalizers

spjmurray:
At present we rely on resources living in an identity managed project namespace in order to get cascading, blocking deletion semantics.  This is bad for a number of reasons: a) the service needs to live in the same cluster, b) the service needs to do a bunch of indirection to find where the resource exists, and has knowledge of identity's layout.

Looking forward, what we want is to seaprate concerns, so identity can do what it wants, and the external service can do what it wants with very little coupling.  Taking compute as an example, when we create a cluster, we want to mark the project as "in use" by the cluster, blocking deletion until we are done with it.  Cluster deletion can either be explicit, or implicit via a message bus (e.g. Kubernetes watch, or a message queue).  In either case, once cleanup has completed we can remove the reference from the project in order to allow deletion.

We use an API, rather than direct project access to enforce data boundaries and reinforce the whole decoupling drive.

As a side effect, for compute in particular, all clusters can be moved to a single namespace, meaning roles can be tightened.  Resource lookups can be done directly by ID, rather than walking a "directory" structure first.  In whole this leads to increased performance, security, and scalability.


spjmurray:
# Testing
## Does creation work idempotently?
```shell
[Wed  3 Sep 10:31:55 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/compute curl -k -H "AUthorization: bearer ${AT}" https://identity.spjmurray.co.uk/api/v1/organizations/b714ef50-7c49-438c-9c1f-770c9a8c6383/projects/fbcf400d-4dd5-4dc0-803a-1c710b0e66cb/references/cluster.compute.unikorn-cloud.org%2F9d298a2f-5943-4b86-9ddf-674cc866f62b -X PUT
[Wed  3 Sep 10:31:58 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/compute curl -k -H "AUthorization: bearer ${AT}" https://identity.spjmurray.co.uk/api/v1/organizations/b714ef50-7c49-438c-9c1f-770c9a8c6383/projects/fbcf400d-4dd5-4dc0-803a-1c710b0e66cb/references/cluster.compute.unikorn-cloud.org%2F9d298a2f-5943-4b86-9ddf-674cc866f62b -X PUT
```

```shell
[Wed  3 Sep 10:30:18 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/identity kubectl get projects.identity.unikorn-cloud.org -n organization-77s9b fbcf400d-4dd5-4dc0-803a-1c710b0e66cb  -o yaml | yq .metadata.finalizers
- unikorn
- cluster.compute.unikorn-cloud.org/9d298a2f-5943-4b86-9ddf-674cc866f62b
[Wed  3 Sep 10:32:09 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/identity kubectl get projects.identity.unikorn-cloud.org -n organization-77s9b fbcf400d-4dd5-4dc0-803a-1c710b0e66cb  -o yaml | yq .metadata.finalizers
- unikorn
- cluster.compute.unikorn-cloud.org/9d298a2f-5943-4b86-9ddf-674cc866f62b
```
*Yes!*

## Does deletion work idempotently?

```shell
[Wed  3 Sep 10:32:11 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/compute curl -k -H "AUthorization: bearer ${AT}" https://identity.spjmurray.co.uk/api/v1/organizations/b714ef50-7c49-438c-9c1f-770c9a8c6383/projects/fbcf400d-4dd5-4dc0-803a-1c710b0e66cb/references/cluster.compute.unikorn-cloud.org%2F9d298a2f-5943-4b86-9ddf-674cc866f62b -X DELETE
[Wed  3 Sep 10:32:18 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/compute curl -k -H "AUthorization: bearer ${AT}" https://identity.spjmurray.co.uk/api/v1/organizations/b714ef50-7c49-438c-9c1f-770c9a8c6383/projects/fbcf400d-4dd5-4dc0-803a-1c710b0e66cb/references/cluster.compute.unikorn-cloud.org%2F9d298a2f-5943-4b86-9ddf-674cc866f62b -X DELETE
```

```shell
[Wed  3 Sep 10:32:12 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/identity kubectl get projects.identity.unikorn-cloud.org -n organization-77s9b fbcf400d-4dd5-4dc0-803a-1c710b0e66cb  -o yaml | yq .metadata.finalizers
- unikorn
[Wed  3 Sep 10:32:19 BST 2025] simon@symphony ~/src/github.com/unikorn-cloud/identity kubectl get projects.identity.unikorn-cloud.org -n organization-77s9b fbcf400d-4dd5-4dc0-803a-1c710b0e66cb  -o yaml | yq .metadata.finalizers
- unikorn
```

*Yes!*

charts/identity/values.yaml
```diff
@@ -150,14 +151,17 @@ roles:
...
         identity:quotas: [create,read,update,delete]
         identity:allocations: [create,read,update,delete]
         region:regions: [create,read,update,delete]
         region:flavors: [create,read,update,delete]
         region:images: [create,read,update,delete]
         region:externalnetworks: [create,read,update,delete]
         region:identities: [create,read,update,delete]
+        region:identities/references: [create,read,update,delete]
         region:networks: [create,read,update,delete]
+        region:networks/references: [create,read,update,delete]
```
spjmurray, line 164: V1 Kubernetes will directly reference the network as all clusters need to reference one.  V1 compute already conforms to this way of doing things.

pkg/client/tokenissuer.go
```diff
@@ -43,6 +43,14 @@ var (
 	ErrResponse = errors.New("unexpected http response")
 )
 
+// ServiceDescriptor is used to define tracing information.
+type ServiceDescriptor struct {
```
spjmurray, line 47: This is primarily a readability thing, especially in the client code that can just do a `version.ServiceDescriptor()` call, rather than passing around multiple values.

spjmurray:
And we're off!  Consult https://github.com/nscaledev/uni-region/pull/184 for client code changes.

squaremo (changes requested):
I can see the substantive change here, which is to add `.../{project}/references/{ref}` to the identity API and the client for the identity API, and to implement the handler which adds a finaliser for each reference. A reasonable motivation is given in the PR description and the code checks out.

There's some changes included that are related but not part of that change:
- rejigging permissions to be more precise
- making the tracing setup a bit cleaner


- [ ] Those should go in their own commits.

The main change has a good explanation in its commit message. There's a couple of things I don't quite get:

> Looking forward, what we want is to separate concerns, so identity can
do what it wants, and the external service can do what it wants with
very little coupling.  Taking compute as an example, when we create a
cluster, we want to mark the project as "in use" by the cluster,
blocking deletion until we are done with it.  Cluster deletion can
either be explicit, or implicit via a message bus (e.g. Kubernetes
watch, or a message queue).  In either case, once cleanup has completed
we can remove the reference from the project in order to allow deletion.

This suggests that deletion events from clusters come through on a message queue. How does the project controller know how to listen on a message queue? I think what it's actually getting at is: here, we don't have to care how cluster deletion works, only that once the cleanup work is done for a cluster, something will remove the reference to the project. (If so, perhaps that could be made a bit clearer).

Second thing: it's not clear why individual references are used, rather than finalizers, since the desired behaviour is that of finalizers. This is pertinent here (and deserves explanation), because it's unorthodox to use Kubernetes finalizers in this way.

Lastly:

- [ ] tests please.

Often in UNI it's hard to test stuff because there are tricky external dependencies. I don't think that's the case here. Tests are evidence that the change is correct, and a guard against it becoming accidentally incorrect later.

spjmurray:
Right pending tests... which will probably not happen until Friday now.  Wednesday at a push.

PR #304, opened 2025-10-02
Machinery for using an external OIDC provider

squaremo:
This adds code in pkg/ for letting services authenticate using an external OIDC provider.

It is not as simple as just directing all requests to the external provider, unfortunately: it's likely that in any deployment of UNI, there will be long-lived service account tokens in the wild -- and new tokens may get created if the service account system is used. So: tokens are dispatched based on whether they are recognisable as issued externally, and otherwise punted to the identity service.

I've put in some convenience funcs for setting this internal-external dispatch up for other services, but stopped short of changing the identity service itself (which after all, would not use the same constructions). Nothing should change in running code.

Once the machinery is merged and addressable, I can introduce its use in all the services, configured through their charts.


squaremo:
I moved this to draft because Matt and I came up with what's potentially a much better scheme:

- implement token exchange in the identity service;
- the token exchange implementation in the ID service dispatches tokens against either internal records, or configured connectors (e.g., Auth0), then mints a new, internal, encrypted token including the subject (and any other info needed for authz? possibly scoped to the organisation), which is then used for all client requests
- the identity service /userinfo only deals with internal tokens, by decrypting them and returning the subject (and other info) therein); this means *it doesn't need to have its own user records*
- the (outermost) middleware used in API servers calls token exchange to get internal tokens, which can then be used for client calls to other services

PR #306, opened 2025-10-06
Bump Core

spjmurray:
Bumps the core library to bring in shared server configuration that aims to simplify things, and make common, operation across all components.


spjmurray:
Note: do not merge in this state, it needs the go.mod reference updating to HEAD once committed... although it'll still work.

PR #307, opened 2025-10-06
Split Up Infrastructure Service Roles

spjmurray:
Compute and Kubernetes service requirements have diverged somewhat, and will only get more out of sync when V2 occurs.  This splits the exising role into two new ones which only have the required permissions for operation.  Where V2 will tighten up the roles further I've added comments about what needs to be deleted.


charts/identity/values.yaml
```diff
@@ -172,32 +172,32 @@ roles:
         compute:clusters: [create,read,update,delete]
         application:applications: [create,read,update,delete]
         application:applicationsets: [create,read,update,delete]
-  # An infrastructure manager service is a role primarily for Kubernetes like
-  # services that can manage identities and physical networks on behalf of a cluster.
-  infra-manager-service:
-    decription: Infrastructure manager service
+  kubernetes-service:
+    decription: Kubernetes service
```
squaremo, line 176: Oops missed this. Fixed in #294

PR #317, opened 2025-10-15
Decouple authorisation from user and membership records

squaremo:
After removing the references from Group objects to OrganizationUser objects (#314), the next step is to rewrite RBAC to rely only on the claims in the userinfo. This breaks the coupling between the RBAC and the user database, so that an external provider can be used.

The custom claims in question are in an optional, namespaced object in the userinfo claims. The identity service populates it for users and service accounts that it holds; platform-specific hooks will need to be used for other providers.

The correctness of the changes is demonstrated by continuing to pass the ACL tests in pkg/rbac/groups_test.go (with the necessary change to provide the custom claims).

~**NB** this is currently targeted at `decouple-rbac-groups` because it depends on the .Subjects field in the Group type. It'll be best to get #314 merged, then retarget this one.~


alansy-nscale:
Rebased: https://github.com/nscaledev/uni-identity/tree/decouple-auth-from-userdb-alt

spjmurray:
@claude review

squaremo:
> @claude review

The app adds eyes, but I don't think there's a workflow triggered. We can probably add a trigger and call the shared review definition.

spjmurray:
Probably needs rebasing anyway so it has the yaml.  And possibly a workflow trigger would be handy.

nscale-peter:
@claude review

squaremo:
Superseded by #433.

PR #357, opened 2025-12-12
Enable organisation-level allocations

alansy-nscale:
## Description

This change introduces support for organisation-level allocations. The new APIs are dedicated to organisation-scoped allocations, while the existing ones remain responsible for project-scoped allocations. This removes the need to infer project IDs from incoming requests and avoids searching for allocations across multiple namespaces, resulting in clearer logic and more predictable behaviour.


pkg/handler/allocations/client.go
```diff
@@ -151,52 +162,40 @@ func generate(ctx context.Context, namespace *corev1.Namespace, organizationID,
...
 
-	return convertList(result), nil
+	return convertList(&list), nil
 }
 
-func (c *SyncClient) Create(ctx context.Context, organizationID, projectID string, request *openapi.AllocationWrite) (*openapi.AllocationRead, error) {
-	namespace, err := common.New(c.client).ProjectNamespace(ctx, organizationID, projectID)
-	if err != nil {
-		return nil, err
+func (c *SyncClient) Create(ctx context.Context, organizationID string, projectID *string, request *openapi.AllocationWrite) (*openapi.AllocationRead, error) {
```
squaremo: This changes to `project *string` so that it can account for both org- and project-level allocations (rather than having two methods, which would be OK too).

PR #362, opened 2025-12-15
Add Image Quota Metadata to Helm Chart Values

alansy-nscale:
## Description

This PR updates the Helm chart values to introduce a new quota metadata for images.


charts/identity/values.yaml
```diff
@@ -343,6 +343,10 @@ quotas:
     displayName: Networks
     description: Layer 3 networks.
     default: 5
+  images:
+    displayName: Images
+    description: Container images.
+    default: 5
```
squaremo: I reckon this should be `0`, so that no-one has a quota unless we assign it.

PR #427, opened 2026-03-19
feat: add object storage permissions for classes and endpoints

nsricardor:
This PR updates the default identity role scopes to support the new S3-compatible object storage offer.

It adds support for:
- storage:objectstorageclasses
- storage:objectstorageendpoints
- storage:objectstorageendpoints/accesskeys

Although access keys are defined per objectstorageendpoint, the dedicated scope scope gives better security granularity by allowing access key operations to be controlled independently from endpoint operations.


squaremo (approved):
Looks reasonable. It does strike me, _en passant_, that
- we need to make this composable, rather than accumulating cruft here
- as with K8s RBAC it's hard to reason about what roles _actually need_
.. but not problems for this change to solve.

PR #429, opened 2026-03-19
feat(ci): KinD-based integration test pipeline

spjmurray:
## Background

Until now the only automated test coverage on pull requests came from
unit tests and contract tests.  These test individual functions and HTTP
contract shapes in isolation, but give no signal that the Helm chart is
correct, that RBAC policies are actually enforced against real HTTP
requests, or that the controller reconciliation loop produces the
expected Kubernetes state.  A single misconfigured Helm value or a
missing RBAC rule can silently slip into main and only surface when a
downstream service deploys.

This commit adds a full end-to-end integration pipeline that runs on
every pull request.  It spins up a fresh KinD cluster, deploys identity
from source, creates shared test fixtures, and runs the existing
test/api suite in randomised, race-detected mode.  The same scripts
double as a local developer workflow so that contributors can reproduce
CI failures on their own machines before pushing.

### GitHub Actions workflow (.github/workflows/integration.yaml)

New job, triggered on PR open / synchronise / reopen / ready_for_review.
Steps in order:

1. Check out the repository.
2. Install Go (version from go.mod), yq, and Helm.
3. Create a KinD cluster named identity-<run_id> using
   hack/ci/kind-config.yaml.
4. make integration-infra  — install MetalLB, cert-manager, ingress-nginx,
   and unikorn-core into the cluster.
5. make integration-install — build images from source, load them into
   KinD, deploy identity via Helm with a random release name / namespace,
   wait for TLS and JOSE-key readiness.
6. make integration-fixtures — issue an mTLS client cert, call the
   identity HTTP API to create an org, two groups, a project, and two
   service accounts (one administrator, one user), write test/.env.
7. make test-api-ci — run test/api in randomised, race-detected mode.
8. On failure, upload junit.xml and test-results.json as a workflow
   artefact for post-mortem.

The cluster is named per run-id to avoid conflicts if multiple jobs run
concurrently on the same runner pool.  A random 8-character suffix seeds
all Helm release names and ingress hostnames so that no test can rely on
a hardcoded name; this has caught real bugs.

### hack/install-metallb/ (new Go tool)

A self-contained Go binary (`go run ./hack/install-metallb`) that installs
MetalLB into a KinD cluster and configures an IP address pool derived
from the cluster's Docker network.

- Applies the upstream metallb-native.yaml manifest.
- Waits for the MetalLB controller Deployment to be Available and the
  speaker DaemonSet to be fully ready.
- Probes the Docker network attached to the KinD control-plane container
  to determine its IPv4 subnet.  For large subnets (≥ /25) it carves a
  .200–.250 range out of the last /8 block; for smaller subnets it falls
  back to the last 51 usable addresses inside the subnet.
- Applies an IPAddressPool and L2Advertisement using the computed range.

This replaces the previous approach of mapping host ports 80/443 into the
cluster via extraPortMappings.  MetalLB assigns a real IP to the
ingress-nginx LoadBalancer Service, which is then used directly as the
nip.io base for ingress hostnames.  This is more faithful to production
and does not require elevated host-port privileges.

`hack/install-metallb/main_test.go` covers the VIP-range calculation logic
with three table-driven cases: large subnet (.200–.250 in the last /8),
small subnet fallback (last 51 usable IPs), and a panic guard for
undersized subnets.

### hack/ci/ — composable install scripts

**Composable install model**: each service defines its own
`hack/ci/install` unit.  Downstream services (e.g. compute) call
identity's install script directly rather than duplicating its logic.
Fix identity's install once; all dependants benefit automatically.

**hack/ci/install** (updated)
  Switched from yq to a pure-awk `chart_version` function so the script
  has no dependency on yq being installed.

  For KinD clusters the ingress hostname is now derived from the actual
  LoadBalancer IP assigned by MetalLB rather than hardcoded `127.0.0.1`:
  ```
  INGRESS_HOST="identity-${SUFFIX}.${INGRESS_IP}.nip.io"
  ```
  Fails fast if ingress-nginx has no LoadBalancer IP yet (surfaces the
  error clearly rather than producing a broken hostname).

  Outputs an additional `IDENTITY_INGRESS_IP` variable in the .env
  fragment so callers can use the IP directly if needed.

  Non-KinD clusters retain the previous `.test` domain suffix.

**hack/ci/setup-infra** (updated)
  Switched KinD ingress from hostPort/NodePort to MetalLB + LoadBalancer:
  - Captures `KIND_CLUSTER_NAME` from the kubectl context (strips `kind-` prefix).
  - Runs `go run ./hack/install-metallb --cluster-name "${KIND_CLUSTER_NAME}"`
    to install MetalLB and configure the address pool before ingress-nginx.
  - Sets `controller.service.type=LoadBalancer` instead of NodePort.
  - Removes the `hostPort.enabled=true` and NodePort-specific Helm flags.
  - After ingress-nginx installs, waits for the LoadBalancer IP to appear
    on the Service (kubectl wait with jsonpath condition, 60 s timeout).

**hack/ci/kind-config.yaml** (updated)
  Removed the `extraPortMappings` block (host ports 80/443).  These were
  required by the old hostPort approach but are no longer needed now that
  MetalLB assigns a routable IP to the LoadBalancer Service.  The
  `ingress-ready=true` node label is still applied.

### pkg/middleware/openapi/openapi.go — ACL cache key scope fix

`aclCacheKey` previously keyed the per-request ACL cache on the caller's
subject (or impersonated actor) alone.  This caused stale cross-scope
reuse: a caller that first hit `/api/v1/acl` (global) and then hit
`/api/v1/organizations/{id}/acl` received the cached global ACL for the
scoped request because both produced the same cache key.

The function now accepts an `organizationID` parameter and appends it (or
`"_global"` when absent) to the key:

```
sub|<organizationID>   — scoped call
sub|_global            — unscoped call
```

The same logic applies to impersonated calls.  `validateRequest` passes
`params["organizationID"]` to `aclCacheKey`, which is empty for routes
that carry no `{organizationID}` path parameter.

### test/api — suite and config changes

**acl_test.go**: Restored the "Given invalid organization ID" test
  together with the ACL cache key fix.  The previous commit omitted this
  test case because the stale-cache bug meant a caller hitting the scoped
  route after the global route received the cached global ACL instead of
  the correctly scoped one — making the test non-deterministic.  With the
  cache key now including the organization scope, the scoped and global
  routes always populate independent cache entries, and the test reliably
  verifies that `acl.Organization` is absent (nil) when the requested
  organization ID is not in the caller's scope.

**suite_test.go**: Constructs `adminClient` and `userClient` in `BeforeEach`.

**config.go**: Added `AdminToken`, `UserToken`, `UserSAID` fields.

**api_client.go**: Added `GetEndpoints()` accessor.

**rbac_matrix_test.go** (new): Verifies RBAC policy via real HTTP
  round-trips.  Uses typed client methods, BDD structure, body assertions
  on administrator paths, and skip-rather-than-fail for missing tokens.

**users_test.go**: Skips rather than fails when the users list is empty
  (integration clusters have no OIDC-synced human users).

### Documentation

**docs/integration-testing.md** (new): Full developer guide covering
  both KinD and Colima workflows, prerequisites, one-time cluster setup,
  iterative workflows, manual mode, troubleshooting, and CI artefacts.

**test/README.md**: Updated to reference the new integration flow.

**test/.env.example**: Updated to include `ADMIN_AUTH_TOKEN`,
  `USER_AUTH_TOKEN`, and `TEST_USER_SA_ID`.

**README.md**: Added "Integration Testing" section.

**CLAUDE.md**: Added integration testing spec URL and Test Writing section.

### .gitignore

Added top-level `junit.xml` and `test-results.json` patterns to cover
artefact files emitted outside the `test/api/suites/` directory.

### DIFF
--- a/test/.env.example
+++ b/test/.env.example
@@ -1,33 +1,34 @@
-# Configuration for Identity API Tests
+# test/.env.example
 #
-# For local testing:
-#   1. Copy this file: cp test/.env.example test/.env
-#   2. Update the values below with your actual configuration
+# Template showing the variables written by `make integration-fixtures`.
+# This file is committed; test/.env is gitignored.
 #
-# Or use environment-specific files if available:
-#   - For dev: cp test/.env.dev test/.env
-#   - For UAT: cp test/.env.uat test/.env
+# Normally test/.env is generated automatically:
+#   make integration-install integration-fixtures   # writes test/.env.install then test/.env
+#   make test-api                                  # reads test/.env
+#
+# To point the suite at an existing deployment without re-running fixtures:
+#   KIND_SUFFIX=<suffix> make integration-fixtures test-api
 
-# Base URL for the Identity API server
-IDENTITY_BASE_URL=http://localhost:8080
+# Identity service URL (set by hack/ci/install)
+IDENTITY_BASE_URL=https://identity-<suffix>.127.0.0.1.nip.io
 
-# Authentication token for API requests
-# This is a service token, which you can create on console, no need to add bearer, it will add it for you
-API_AUTH_TOKEN=your-test-token-here
+# Absolute path to the CA bundle written by hack/ci/setup-infra
+IDENTITY_CA_CERT=/path/to/hack/ci/ca-bundle.pem
 
-# Test data configuration
-# Advised you create all the below in console first and add them all here
-TEST_ORG_ID=test-org-12345
-TEST_PROJECT_ID=test-project-67890
+# Resource IDs created by hack/ci/fixtures
+TEST_ORG_ID=
+TEST_PROJECT_ID=
+TEST_ADMIN_GROUP_ID=
+TEST_USER_GROUP_ID=
+TEST_ADMIN_SA_ID=
+TEST_USER_SA_ID=
 
-# Test timeout settings
-REQUEST_TIMEOUT=30s
-TEST_TIMEOUT=20m
+# Bearer token for the administrator service account
+# (organization-scoped; full identity CRUD)
+API_AUTH_TOKEN=
+ADMIN_AUTH_TOKEN=
 
-# Test behavior flags
-SKIP_INTEGRATION=false
-
-# Debug settings
-DEBUG_LOGGING=false
-LOG_REQUESTS=true
-LOG_RESPONSES=false
+# Bearer token for the user service account
+# (project-scoped; limited org-level access)
+USER_AUTH_TOKEN=
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -1,0 +1,72 @@
+name: Integration
+on:
+  pull_request:
+    types:
+    - opened
+    - synchronize
+    - reopened
+    - ready_for_review
+
+permissions:
+  contents: read
+
+jobs:
+  Integration:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
+    - name: Setup Go
+      uses: actions/setup-go@v6
+      with:
+        go-version-file: go.mod
+
+    - name: Install yq
+      run: |
+        wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
+        chmod +x /usr/local/bin/yq
+
+    - name: Install Helm
+      uses: azure/setup-helm@v4
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Generate random suffix
+      id: rand
+      run: echo "suffix=$(head /dev/urandom | tr -dc a-z0-9 | head -c 8)" >> "$GITHUB_OUTPUT"
+
+    - name: Create KinD cluster
+      uses: helm/kind-action@v1
+      with:
+        cluster_name: identity-${{ github.run_id }}
+        config: hack/ci/kind-config.yaml
+
+    - name: Install cluster infrastructure
+      env:
+        KIND_CLUSTER: identity-${{ github.run_id }}
+      run: make integration-infra
+
+    - name: Install identity
+      env:
+        KIND_CLUSTER: identity-${{ github.run_id }}
+        KIND_SUFFIX: ${{ steps.rand.outputs.suffix }}
+      run: make integration-install
+
+    - name: Create test fixtures
+      env:
+        KIND_CLUSTER: identity-${{ github.run_id }}
+        KIND_SUFFIX: ${{ steps.rand.outputs.suffix }}
+      run: make integration-fixtures
+
+    - name: Run API integration tests
+      run: make test-api-ci
+
+    - name: Upload test artifacts
+      if: failure()
+      uses: actions/upload-artifact@v4
+      with:
+        name: api-test-results
+        path: |
+          junit.xml
+          test-results.json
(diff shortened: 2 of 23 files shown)


JakeNScale (commented):
For the new tests themselves, consistency with the other patterns in api and contract would be good.
Some things that we are doing in those suites and not here are:

- Using Ginkgo + Gomega with BDD nesting: Describe > Context > Describe > It
- Typed API client (api.APIClient) instead of raw http.Client calls, there are actually quite a few methods we can use from api_client that e2e can just import and use rather than re-using, like endpoints.go & config.go
- Rich assertions on response bodies (metadata, IDs, status fields), instead of just status codes

Also I think we would want to expand the e2e a bit more, which we can do in subsequent tickets, for example create → get → update → delete sequence. But that can come after, as this PR's primary focus is standing up the framework.

test/.env.example
```diff
@@ -1,33 +1,33 @@
...
+# This file is committed; test/.env is gitignored.
 #
-# Or use environment-specific files if available:
-#   - For dev: cp test/.env.dev test/.env
-#   - For UAT: cp test/.env.uat test/.env
-
-# Base URL for the Identity API server
-IDENTITY_BASE_URL=http://localhost:8080
+# Normally test/.env is generated automatically:
+#   make kind-install kind-fixtures   # writes test/.env.install then test/.env
```
JakeNScale: ❤️

spjmurray:
I will see if Claudette can infer some test rules from existing tests and populate AGENTS.md with the necessary pointers.

spjmurray:
In your face Claude!
```
● The substantive concern is from Jake: BDD nesting, typed client, and body assertions — all of which    
  were addressed in the last commit. Let me assess Claude bot's specific criticisms:                     
                                                                                                         
  "Hardcoded URLs in DescribeTable entries" — no weight. Claude suggests endpoints.ListGroups("{orgID}") 
  but Endpoints.ListGroups calls url.PathEscape("{orgID}"), which produces %7BorgID%7D. The              
  strings.ReplaceAll substitution would then fail to find {orgID}. The string literals are necessary     
  here.                                                                                                  
                                                                                                         
  "doRawRequest is raw HTTP" — no weight. It's explicitly for RBAC denial checks (403 responses). The    
  typed client's doRequest returns an error on non-200, so it can't be used for expected-failure cases.  
  CLAUDE.md explicitly permits status-only assertions for RBAC tests.                                    
                                                                                                         
  "Error message loses response body" — factually wrong. client.go already includes the body:            
  fmt.Errorf("unexpected status %d for %s %s: %s", resp.StatusCode, method, path, string(body)).         
                                                                                                         
  "TEST_ADMIN_GROUP_ID missing from required check" — minor and intentional; those fields aren't used by 
  any current test, so leaving them optional is correct.                                                 
                                                                                                         
  "DescribeTable inconsistent with BDD nesting" — no weight. DescribeTable is idiomatic Ginkgo for matrix
   tests and is wrapped in Context blocks anyway. Forcing it into It blocks would lose the table-driven  
  structure entirely.                                                                                    
                                                                                                         
  In short: Jake's original concerns are addressed, Claude bot found nothing new of substance. The PR is 
  good to go.
```

JakeNScale (approved):

spjmurray:
Not so fast Jesus... I have discovered our friend Claudette didn't listen to me, I wanted the _existing_ tests to be compatible with this workflow and not invent something new.

spjmurray:
Now we're good @JakeNScale

spjmurray:
Now we're genuinely good, I had to tweak the network model to get region working in concert with identity.

JakeNScale (approved):

PR #434, opened 2026-04-01
feat: type-safe organization and project ID path parameters

spjmurray:
Introduces two distinct UUID-backed types for organization and project
path parameter IDs:

  type OrganizationID uuid.UUID
  type ProjectID      uuid.UUID

Wired into the OpenAPI path parameter schemas via x-go-type with
format: uuid, delivering two orthogonal safety properties:

Runtime safety: OrganizationID and ProjectID implement
encoding.TextUnmarshaler via uuid.UUID, so the oapi-codegen parameter
binder validates UUID format before any handler is reached. Non-UUID
path values produce a 400 at the binding layer rather than propagating
into business logic. This is verified by integration tests that assert
non-UUID organization and project ID path parameters are rejected with
400 before the handler is invoked.

Compile-time safety: the two types are distinct named types, not
aliases, so the compiler prevents a ProjectID from being passed where
an OrganizationID is expected and vice versa. The uuid.UUID underlying
type also enables direct use with APIs such as
GenerateDeterministicResourceID without an intermediate string parse.

Backwards compatibility: the generated ClientWithResponsesInterface now
carries typed OrganizationID and ProjectID parameters. Downstream
consumers update their mocks by regenerating them; the only breaking
change is the two method signatures that carry org or project IDs.

The conversion seams are intentionally shallow. Going inward, .String()
converts typed IDs to strings at the first callee boundary inside each
handler body, leaving subpackage clients unchanged. Going outward,
Parse* converts string-typed IDs from Kubernetes labels or token claims
into typed form at the point they enter the generated API client,
propagating parse errors rather than panicking on malformed input.

AllowProjectScopeCreate now returns OAuth2InvalidRequest (400) rather
than OAuth2AccessDenied (403) when the org or project ID supplied in a
v2 request body is not a valid UUID — invalid input is a bad request,
not an authorisation failure.

An existing integration test that used "invalid-project-id" as a
shortcut for "project not found" has been updated to use a real UUID,
since UUID validation now fires first and returns 400 before the
not-found path is ever reached.

Test constants updated to valid RFC4122 UUIDs to match the contract
the types now enforce.


spjmurray:
Okay, the core change is in https://github.com/nscaledev/uni-core/pull/280.  This is the largest sadly as identity has more API surface area, but you've seen one handler you've seen them all...

pkg/openapi/server.spec.yaml
```diff
@@ -1152,49 +1152,49 @@ components:
...
-        type: string
+        $ref: 'https://raw.githubusercontent.com/nscaledev/uni-core/id_validation/pkg/openapi/common.spec.yaml#/components/schemas/resourceId'
     allocationIDParameter:
       name: allocationID
       in: path
       description: A resource allocation ID.
       required: true
       schema:
-        type: string
+        $ref: 'https://raw.githubusercontent.com/nscaledev/uni-core/id_validation/pkg/openapi/common.spec.yaml#/components/schemas/resourceId'
```
spjmurray: Parameters have been updated, but I'm fairly certain some thing e.g. groups reference users and role IDs, these need strict types.

pkg/handler/allocations/client.go
```diff
@@ -153,8 +160,8 @@ func (c *Client) get(ctx context.Context, namespace, allocationID string) (*unik
 	return result, nil
 }
 
-func (c *SyncClient) Create(ctx context.Context, organizationID, projectID string, request *openapi.AllocationWrite) (*openapi.AllocationRead, error) {
-	namespace, err := common.New(c.client).ProjectNamespace(ctx, organizationID, projectID)
+func (c *SyncClient) Create(ctx context.Context, organizationID, projectID uuid.UUID, request *openapi.AllocationWrite) (*openapi.AllocationRead, error) {
+	namespace, err := common.New(c.client).ProjectNamespace(ctx, organizationID.String(), projectID.String())
```
spjmurray: I don't like the String() calls here, update the common interfaces to push this type conversion lower down the stack.

spjmurray:
Thanks. I agree with the general architectural read and with the point that direct tests for the new nominal ID helpers would be useful.

I do not think the two stated blockers hold up as written.

1. The incomplete parameter typing concern is too speculative. At this point it needs a concrete example of a path/query parameter that is still generated from an inline string schema rather than a typed ID schema. Without that, I do not think this is actionable review feedback.

2. The remaining `.String()` calls are now mostly at genuine string boundaries: Kubernetes object keys / metadata names, token subject strings, and self-access comparison against `userinfo.Sub`. That is exactly where the conversion belongs. Wrapping those in tiny `objectKeyForX` helpers would mostly be stylistic churn rather than a meaningful type-safety improvement.

So my read is: the broad direction is right, the implementation is materially cleaner than the review suggests, and the remaining concerns need either concrete examples or a lower severity classification.

spjmurray:
It's like Robot Wars!

spjmurray:
## Follow-up: Verification of Raised Concerns

### 1. Incomplete Type Enforcement — Not an issue

All path parameters in `server.spec.yaml` use `$ref` to typed schemas, not inline `type: string`:

| Parameter | Schema |
|---|---|
| `organizationIDParameter` | `$ref: organizationId` |
| `serviceAccountIDParameter` | `$ref: serviceAccountId` |
| `userIDParameter` | `$ref: userId` |
| `projectIDParameter` | `$ref: projectId` |
| `groupidParameter` | `$ref: groupId` |
| `oauth2ProvderIDParameter` | `$ref: oauth2ProviderId` |
| `allocationIDParameter` | `$ref: allocationId` |

There is no `roleIDParameter` because roles are never path parameters — they only appear as arrays within group specs. The two remaining string-typed parameters (`referenceParameter`, `userEmailParameter`) are correctly typed as strings since they are not UUIDs. This concern is unfounded.

### 2. String() Calls in Wrong Layer — Not an issue

The `.String()` calls in handlers are at the correct layer: the boundary between the typed API world and Kubernetes, which uses raw strings. The implementation is consistent and well-structured:

- `common.IDStrings[T]` provides a generic bulk converter for slices
- `common.ContainsIDString[T]` handles typed search against Kubernetes string values
- `organizationSelectorUUID` / `projectSelectorUUID` are adapter functions that accept typed IDs and delegate to string-based selector builders

The remaining inline `.String()` calls (e.g. `client.ObjectKey{Name: groupID.String()}`) are single-line, obvious conversions at the point of use — no different from calling `.Error()` on an error. The reviewer's proposed alternative (a dedicated helper per resource type wrapping one line of code) would add indirection with no benefit. Nothing needs changing here.

pkg/handler/handler.go
```diff
@@ -323,12 +323,12 @@ func (h *Handler) GetApiV1OrganizationsOrganizationIDAcl(w http.ResponseWriter,
 }
 
 func (h *Handler) GetApiV1OrganizationsOrganizationIDRoles(w http.ResponseWriter, r *http.Request, organizationID openapi.OrganizationIDParameter) {
-	if err := rbac.AllowOrganizationScope(r.Context(), "identity:roles", openapi.Read, organizationID); err != nil {
+	if err := rbac.AllowOrganizationScope(r.Context(), "identity:roles", openapi.Read, organizationID.String()); err != nil {
```
spjmurray: It doesn't seem like there's a good reason not to propagate type safety lower here, we will need some new RBAC interfaces e.g. `AllowOrganizationScopeTyped` or `TypedAllowOrganizationScope` so that it's backward compatible for other repos including these calls.

pkg/handler/handler.go
```diff
@@ -323,12 +323,12 @@ func (h *Handler) GetApiV1OrganizationsOrganizationIDAcl(w http.ResponseWriter,
...
 		errors.HandleError(w, r, err)
 		return
 	}
 
-	result, err := roles.New(h.client, h.namespace).List(r.Context(), organizationID)
+	result, err := roles.New(h.client, h.namespace).List(r.Context(), organizationID.String())
```
spjmurray: roles however is completely internal to this repo and can be changed to propagate the organizationID for longer without having to create a backwards compatible interface.

spjmurray (commented):
On the whole a good start, it's rather nice how you have been able to keep the clients working (the line protocol is the same it's just the parsing that's different), so you just need to bump the library dependency.

spjmurray:
Note to reviewers, there were 2 separate issues.  The first is that we don't validate the shape of the input (this one) so a user could try doing POST /api/v2/networks with projectID of "my-project" and it would get deep into the system.  The other one was that an admin level account was blindly trusting the project without actually seeing if it existed and persisting the badness to the database, that's already been fixed.

spjmurray:
Another note to reviewers: we deliberately keep the scope narrow here, and only worry about the API handlers.  We can extend this transparently across the CRD boundary eventually for full stack coverage (everything just marshals to a JSON string...)

alansy-nscale (approved):
LGTM. I ran a fable style code review on this PR and the findings are below.
```json
[
  {
    "file": "pkg/handler/handler.go",
    "line": 610,
    "category": "EFFICIENCY",
    "severity": "warning",
    "summary": "organizationID.String() is now called inside the slices.DeleteFunc RBAC filter closure, recomputing the UUID->string format once per project in the list-projects request path; on main organizationID was already a string so the call was free.",
    "failure_scenario": "No build/test break — runtime inefficiency, confirmed by `go build ./pkg/handler/...` passing. In GetApiV1OrganizationsOrganizationIDProjects the closure passed to slices.DeleteFunc runs once per project, and each invocation calls organizationID.String() (formats 16 bytes into a fresh 36-byte string). The organization ID is loop-invariant, so for an org with N projects this performs N redundant UUID formats + N allocations per request; on main organizationID was a plain string passed through with zero per-element cost. Surgical fix: hoist it above the closure — `orgID := organizationID.String()` after `ctx := r.Context()`, then reference orgID inside the DeleteFunc. No second-order consequences: AllowProjectScope already takes a string and orgID is plainly loop-invariant."
  },
  {
    "file": "pkg/handler/projects/client.go",
    "line": 81,
    "category": "DOC",
    "severity": "nit",
    "summary": "google/uuid UnmarshalText (used by typed path params) and ids.Parse* accept non-canonical UUID forms (uppercase, {braces}, urn:uuid:, no-hyphens) and normalize them all to lowercase canonical before they become label-selector values via .String(); this silently widens what the API accepts versus the prior verbatim string match, and is undocumented.",
    "failure_scenario": "Previously a path segment was used verbatim as the Kubernetes label value in organizationSelector/projectSelector, so an uppercase UUID would not match a canonical lowercase label and returned 404. Now organizationID/projectID are parsed by UnmarshalText (verified to accept \"F47AC10B-...\", \"{...}\", \"urn:uuid:...\", and unhyphenated forms, all normalized to canonical lowercase) and the handler passes .String() into the selector — so those previously-404 inputs now resolve to the real resource. This is a widening (more lenient lookups), not a break of any valid flow, since system-generated IDs are already canonical. No code change required; the gap is that pkg/ids/README.md and the parameter descriptions call the value a 'uuid' without noting that several non-canonical encodings are silently accepted and normalized. Surgical fix: add one sentence to pkg/ids/README.md noting UnmarshalText/Parse accept any RFC-4122 textual encoding and canonicalize to lowercase hyphenated form. Second-order: none — .String() round-trips canonical UUIDs unchanged."
  },
  {
    "file": "pkg/ids/README.md",
    "line": 31,
    "category": "DOC",
    "severity": "warning",
    "summary": "The README documents MustParse* as the tool to use 'only where the value is guaranteed valid by a prior validation step (e.g. a path parameter that has already passed the binding layer)'. No such call site exists or can exist: path parameters arrive at handlers ALREADY typed (the binding layer produced the typed ID via UnmarshalText), so the documented canonical use case for MustParse* is unreachable by construction.",
    "failure_scenario": "A new starter reading this README will reach for MustParseProjectID(pathParam) in a handler and discover the handler already receives an ids.ProjectID, or worse will re-stringify-then-MustParse a value, reintroducing a panic path the type system was meant to remove. Fix: rewrite the Conversion section to state MustParse* is a test-only convenience (matching reality) and document the real inward path — untrusted strings from labels/claims/principals go through Parse* with error handling, as references.go and allocations.go actually do."
  },
  {
    "file": "pkg/ids/ids.go",
    "line": 108,
    "category": "SIMPLIFY",
    "severity": "nit",
    "summary": "The unmarshalUUID helper is unnecessary indirection: google/uuid's *UUID.UnmarshalText already does exactly what it does (parse, return on error without mutating, assign only on success), so each UnmarshalText method can call it directly.",
    "failure_scenario": "No build/test break today, but the helper adds a 12-line function and an extra call hop that obscures intent and gives a false impression it provides an atomic-on-error guarantee that the embedded uuid.UUID.UnmarshalText already provides (verified in google/uuid@v1.6.0/marshal.go). Surgical fix: delete unmarshalUUID (ids.go:107-118) and change each of the 7 pointer-receiver methods from `return unmarshalUUID((*uuid.UUID)(v), b)` to `return (*uuid.UUID)(v).UnmarshalText(b)`. Verified: `go build ./pkg/ids/` (exit 0), `go test ./pkg/ids/` (ok), and a probe confirmed the error path still leaves the target zero-valued. Second-order: the `//nolint:recvcheck` directives and value/pointer receiver split must stay (String/MarshalText remain value receivers for fmt.Stringer); no other consequence."
  },
  {
    "file": "pkg/ids/ids.go",
    "line": 120,
    "category": "SIMPLIFY",
    "severity": "nit",
    "summary": "The 14 free functions Parse*ID (×7) and MustParse*ID (×7) each repeat an identical uuid.Parse/uuid.MustParse + named-type-conversion body; because every ID type is backed by uuid.UUID ([16]byte), they collapse to two generic functions Parse[T ~[16]byte] and MustParse[T ~[16]byte].",
    "failure_scenario": "Not a build/test break — code compiles and tests pass. The cost is maintenance duplication: adding an 8th ID type forces hand-copying two more near-identical functions, and a copy-paste slip (e.g. ParseGroupID returning a UserID zero value on error) would not be caught by the compiler. Surgical fix: replace the seven ParseXID bodies with `func Parse[T ~[16]byte](s string) (T, error)` (verified to compile against google/uuid v1.6.0 and Go 1.25) and the seven MustParseXID with `func MustParse[T ~[16]byte](s string) T`; call sites become ids.Parse[ids.OrganizationID](s). Tradeoffs before applying: (1) the per-type String/MarshalText/UnmarshalText method blocks CANNOT be collapsed (Go forbids generic methods and oapi-codegen needs each named type to implement TextUnmarshaler), so this removes ~110 lines, not the type blocks; (2) call sites lose the greppable named form and must spell the type argument; (3) callers in authorizer.go, rbac/handler.go and the *_test.go files need updating. No established convention for this in core/sibling repos, so it is a judgment call — current explicit form is acceptable; flagged only for the duplication-vs-readability decision."
  },
  {
    "file": "pkg/ids/ids_test.go",
    "line": 27,
    "category": "TEST",
    "severity": "warning",
    "summary": "The canonicalization round-trip — the single highest-risk behavior of this change — is never asserted; every Parse/String/UnmarshalText test uses validUUID which is already canonical lowercase, so a regression that stopped canonicalizing (or canonicalized differently) would pass these tests.",
    "failure_scenario": "uuid.UnmarshalText accepts and canonicalizes uppercase ('F47AC10B-...'), URN, braced, and no-hyphen 32-hex forms, all collapsing to the canonical 8-4-4-4-12 lowercase that RBAC scopes and Kubernetes labels are stored in. This canonicalization is exactly what makes RBAC matching (organization.Id != organizationID) and projectSelector/organizationSelector label lookups resolve to the right resource for a non-canonical path input. If a future refactor (e.g. a String() that no longer normalizes, or a custom parser) silently broke normalization, an uppercase-UUID request would start mismatching the canonical ACL scope and flip from allow to 403, or miss the namespace label-selector and 404 — and no unit test would catch it. Surgical fix: add a case asserting ParseOrganizationID(\"F47AC10B-58CC-4372-A567-0E02B2C3D479\").String() == \"f47ac10b-58cc-4372-a567-0e02b2c3d479\" (and the no-hyphen / URN forms), encoding WHY: input casing/format must normalize to the storage form so RBAC and label selectors match. No production code change."
  },
  {
    "file": "pkg/openapi/router.go",
    "line": 625,
    "category": "REGRESSION",
    "severity": "nit",
    "summary": "Malformed-UUID path params now return 400 BEFORE the auth middleware runs, so the previously uniform pre-auth response (401 for any ID shape) now differs by UUID syntax validity — a behavior shift relative to the secure-by-default disclosure model documented in pkg/handler/README.md (lines 124-134).",
    "failure_scenario": "Before this change, every path-param ID was a plain string that always bound successfully; an unauthenticated request to e.g. GET /api/v1/organizations/<id>/projects/<id> always reached the validator middleware and was rejected with 401 regardless of the ID's shape. After this change the generated wrapper binds path params (BindStyledParameterWithOptions, router.go:625-629) and returns 400 via ErrorHandlerFunc->HandleError BEFORE siw.HandlerMiddlewares (audit + validator/authentication) are applied (router.go:641-643). Net effect: an unauthenticated caller can distinguish a malformed-UUID ID (400) from a well-formed one (401) without credentials. This only leaks UUID-syntax validity (already published in the OpenAPI spec) and reveals nothing about resource existence or tenant layout, so it is not an enumeration/policy oracle and is consistent with the documented '400 means request shape wrong' taxonomy. No code fix needed; flagging only because it is a genuine pre-auth response-code change that a reviewer of the disclosure model should be aware of. Confirmed via router.go ordering and pkg/handler/error.go (bind error -> OAuth2InvalidRequest, HTTP 400)."
  },
  {
    "file": "pkg/rbac/handler.go",
    "line": 249,
    "category": "EFFICIENCY",
    "severity": "warning",
    "summary": "AllowRole now calls AllowOrganizationScopeID (which does organizationID.String()) inside its organization-scope and project-scope endpoint/operation loops, recomputing the same UUID->string on every iteration; previously organizationID was a string passed directly.",
    "failure_scenario": "No build/test break (confirmed by `go build ./pkg/rbac/...`) — runtime inefficiency. AllowRole iterates role.Spec.Scopes.Organization and role.Spec.Scopes.Project, and for every (endpoint, operation) pair calls AllowOrganizationScopeID, whose only body is `AllowOrganizationScope(ctx, ..., organizationID.String())`. organizationID is invariant across the function, so a role with E endpoints x O operations triggers E*O redundant UUID formats + allocations. Amplified because AllowRole is itself invoked per element in two list/validate paths: roles.Client.List (slices.DeleteFunc over every role) and groups.validateRoleIDs (loop over every requested role ID). Net cost on a roles-list request is roughly len(roles) x (orgEndpoints+projEndpoints) x ops String() calls, where on main it was zero. Surgical fix: compute `orgID := organizationID.String()` once at the top of AllowRole and call the existing string-typed AllowOrganizationScope in both loops. Second-order: keeps AllowRole's typed signature for callers; the two typed call sites are unaffected."
  },
  {
    "file": "test/api/suites/id_validation_test.go",
    "line": 30,
    "category": "TEST",
    "severity": "warning",
    "summary": "The integration suite exercises only the handler-facing path params, but the branch introduced two NEW server-side string->typed parse points that gate authorization and can now reject with an error that did not exist before: pkg/middleware/openapi/remote/authorizer.go:350 (GetACL parses organizationID) and pkg/rbac/handler.go:209 (AllowProjectScopeCreate parses org/project IDs from cross-repo string callers, returning OAuth2InvalidRequest). Neither new error path is covered by any test.",
    "failure_scenario": "A regression that broke the new parse-then-error logic in the remote authorizer or AllowProjectScopeCreate (e.g. swallowing the parse error, or parsing the wrong variable) would pass the entire test suite, because the only tests touching those functions feed pre-validated UUID constants. Given these functions sit on the authorization path for cross-repo callers, a silent failure here is a security-relevant gap. Fix: add a unit test for AllowProjectScopeCreate with a malformed organizationID/projectID asserting it returns invalid_request (not access_denied), and a unit test for Authorizer.GetACL asserting a non-UUID organizationID errors before any HTTP call."
  }
]
```

spjmurray:
You are a sick man @alansy-nscale for thinking I'm going to read raw JSON...

alansy-nscale:
> Malformed-UUID path params now return 400 BEFORE the auth middleware runs, so the previously uniform pre-auth response (401 for any ID shape) now differs by UUID syntax validity — a behavior shift relative to the secure-by-default disclosure model documented in pkg/handler/README.md (lines 124-134)

The only one that's really worth looking at is just this one. The others are just test gaps and a few nitpicks.

spjmurray:
You could argue rejecting before invoking cryptography is saving our precious CPU cycles....

It did notice URNs... there's a school of thought we could shift to `urn:uni:production:us-east-1:server:a8f29840-ac67-4730-93c4-a6b446cae760`, so it's correctly guessed the possible direction of travel.

PR #435, opened 2026-04-08
fix: resolve group subject issuer inconsistency between users and groups API paths

alansy-nscale:
## Description

This PR fixes a bug where `updateGroups` in the users handler constructed `GroupSubject` with `Issuer: ""` for local users, while `userIDsToSubjects` in the groups handler wrote `Issuer: c.issuer.URL`. The fix is a one-line change in `updateGroups` to use `c.issuer.URL` consistently.

Two integration tests are included to prove both failure modes and serve as regression guards. The first proves the issuer inconsistency directly between the two write paths. The second proves that when the groups API last set membership, deleting the user via the users API leaves a dangling subject in the group: `removeFromGroup` matches by both ID and Issuer, so the `Issuer: ""` constructed at delete time never matches the stored `Issuer: c.issuer.URL`. The second test also documents that this dangling subject bypasses the `subjectsToUserIDs` "Skip external subjects" guard, meaning any subsequent GET→PUT round-trip on the group will hit `findUserBySubject` for the deleted user and return an error.

⚠️ **Patch required**: existing group resources in all clusters will need a one-time patch to backfill the correct issuer URL into any `spec.subjects[].issuer` field that is currently empty or incorrect. Without this patch, the delete and round-trip bugs described above will persist for any group membership that was last written by the old users API path.


alansy-nscale:
@claude review

squaremo (approved):
Thanks for resolving this gap, Alan 🍎

nscale-peter (approved):

spjmurray (approved):

PR #438, opened 2026-04-11
feat(ci): enable KinD integration tests on macOS with Docker Desktop

peterj-ns:
macOS runs Docker inside a Linux VM, so MetalLB-assigned IPs are only routable within the VM and unreachable from the host. This adds Docker Desktop support to the KinD integration test pipeline, routing ingress via hostPort instead.

## Changes

- `kind-config.yaml`: expose container port 443 on host port 8443 via extraPortMappings
- `setup-infra`: detect Docker Desktop, skip MetalLB, use hostPort ingress
- `install`: use 127.0.0.1:8443 as ingress endpoint and construct BASE_URL accordingly
- `suite_test.go`: call `patchTLSTransport()` in BeforeSuite
- `tls_darwin_test.go`: inject self-signed CA into `http.DefaultTransport` (macOS only — Go ignores `SSL_CERT_FILE` in favour of the system Keychain)
- `tls_other_test.go`: no-op `patchTLSTransport` for non-darwin platforms
- `Makefile`: use `$(GOBIN)/oapi-codegen`, simplify `KIND_SUFFIX` via `openssl rand -hex 4` (fixes SIGPIPE on macOS), add `$(OPENAPI_FILES)` to touch target
- `integration.yaml`: align random suffix generation with Makefile (`openssl rand -hex 4`)
- `docs/integration-testing.md`: document Docker Desktop support
- `hack/ci/README.md`: update output contract to reflect Docker Desktop vs MetalLB URL variants

Linux CI is unaffected — `IS_DOCKER_DESKTOP` is false on `ubuntu-latest`, MetalLB path is unchanged.

Closes INST-796


peterj-ns:
superseded by https://github.com/nscaledev/uni-identity/pull/444

PR #439, opened 2026-04-11
test(api): expand integration test coverage for service accounts, users, and projects

peterj-ns:
Expands integration test coverage from list-only to full CRUD for all major identity resources.

## Infrastructure

- `test/api`: typed client methods for create/update/delete on service accounts, users, projects, organizations, and quotas
- `test/api`: `CreateServiceAccountWithCleanup`, `CreateUserWithCleanup`, `CreateProjectWithCleanup` fixtures and payload builders

## New test coverage

- **Service accounts**: create returns one-time `AccessToken`; token absent on update; rotate returns a different token; group membership create/update/clear via list verification; async delete polling
- **Users**: create with subject and state; `Spec.State` asserted directly on create/update response body; group membership update/clear via list verification; suspend/reactivate; delete
- **Projects**: async create (202) with `WaitForProjectProvisioned`; group associations; update after provisioned; async delete via `GetProject` → `ErrResourceNotFound` polling
- **Organizations**: update name with deferred restore
- **Quotas**: `SetQuotas` denied for organization administrator (platform-administrator only)
- **Groups**: SA assignment via `UpdateGroup`; bidirectional membership reflection between groups, users, and service accounts
- **ACL**: all three project membership tests create their own projects via `adminClient` + `config.UserGroupID`; appearance and removal both use `Eventually` + `func() bool` polling to handle propagation lag
- **OAuth2 providers**: create, update, delete, list
- **RBAC matrix**: mutation denial for regular users (create/update/delete group, create SA, list users, set quotas, update org, delete another principal's SA, create/delete OAuth2 provider); user cannot access a project they are not a member of (GET denied, absent from list); admin permitted for list groups/roles/SAs/users/projects

## Notes

- `Spec.GroupIDs` not included in user/SA create or update response bodies — verified via list
- `GET /serviceaccounts/{id}` not exposed by server — list used for SA group membership and delete verification

Relates to INST-797.

### DIFF
--- a/test/api/suites/groups_test.go
+++ b/test/api/suites/groups_test.go
@@ -22,7 +22,6 @@ package suites
 
 import (
 	"errors"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -36,7 +35,7 @@ var _ = Describe("Group Management", func() {
 	Context("When creating groups", func() {
 		Describe("Given valid group data", func() {
 			It("should create a new group with complete metadata", func() {
-				payload := api.NewGroupPayload().BuildTyped()
+				payload := api.NewGroupPayload().Build()
 				group, groupID := api.CreateGroupWithCleanup(client, ctx, config, payload)
 
 				Expect(groupID).NotTo(BeEmpty(), "Group ID should be returned")
@@ -66,7 +65,7 @@ var _ = Describe("Group Management", func() {
 			})
 
 			It("should create a group and retrieve it by ID", func() {
-				payload := api.NewGroupPayload().BuildTyped()
+				payload := api.NewGroupPayload().Build()
 				createdGroup, groupID := api.CreateGroupWithCleanup(client, ctx, config, payload)
 
 				retrievedGroup, err := client.GetGroup(ctx, config.OrgID, groupID)
@@ -81,7 +80,7 @@ var _ = Describe("Group Management", func() {
 			})
 
 			It("should create a group and find it in the organization list", func() {
-				payload := api.NewGroupPayload().BuildTyped()
+				payload := api.NewGroupPayload().Build()
 				createdGroup, groupID := api.CreateGroupWithCleanup(client, ctx, config, payload)
 
 				groups, err := client.ListGroups(ctx, config.OrgID)
@@ -101,11 +100,12 @@ var _ = Describe("Group Management", func() {
 
 				Expect(found).To(BeTrue(), "Created group should appear in organization group list")
 			})
+
 		})
 
 		Describe("Given invalid organization ID", func() {
 			It("should return error when creating group in non-existent organization", func() {
-				payload := api.NewGroupPayload().BuildTyped()
+				payload := api.NewGroupPayload().Build()
 
 				_, err := client.CreateGroup(ctx, "invalid-org-id", payload)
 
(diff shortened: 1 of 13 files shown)


test/api/suites/groups_test.go
```diff
@@ -201,6 +204,102 @@ var _ = Describe("Group Management", func() {
 					originalGroup.Metadata.Name, updatedGroup.Metadata.Name)
 			})
 
+			It("should create a group with user membership", func() {
```
JakeNScale: I don't think this one belongs in the `Given existing group` context, maybe better placed in the `When creating groups`

test/api/suites/groups_test.go
```diff
@@ -201,6 +204,102 @@ var _ = Describe("Group Management", func() {
...
+
+				retrieved, err := client.GetGroup(ctx, config.OrgID, groupID)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(retrieved.Spec.UserIDs).NotTo(BeNil())
+				Expect(*retrieved.Spec.UserIDs).To(ContainElement(userID),
+					"Group should contain the assigned user")
+			})
+
+			It("should reflect group membership in the service account when group is created with a service account", func() {
```
JakeNScale: Same here, I think this is a create block, as apposed to update

test/api/suites/groups_test.go
```diff
@@ -201,6 +204,102 @@ var _ = Describe("Group Management", func() {
...
+						break
+					}
+				}
+
+				Expect(found).NotTo(BeNil(), "Service account should appear in list")
+				Expect(found.Spec.GroupIDs).To(ContainElement(groupID),
+					"Service account should reflect the group assigned via group creation")
+			})
+
+			It("should reflect group membership in the user when group is created with a user", func() {
```
JakeNScale: again, same as above

JakeNScale (commented):
Couple of feedbacks from conductor:

Duplicate test in acl_test.go
The test "should include a newly created project in the organization ACL" (line ~199) is functionally identical to "should include a project the user has group membership on in the organization ACL" (line ~110). Both:

Create a project with config.UserGroupID
Wait for provisioning
Poll ACL until the project appears
The only difference is the description and some debug logging. One of these should be removed.

groups_test.go cross-resource assertions
The two It blocks at lines ~224 and ~252 (the misplaced ones under “When updating groups”) assert on other resources’ state — one calls ListServiceAccounts and checks Spec.GroupIDs on the SA, the other calls ListUsers and checks Spec.GroupIDs on the user. These are really testing “does the SA/user reflect group creation?” — service account and user behavior, not group behavior.

The mirror pattern already exists in the correct locations: serviceaccounts_test.go has "should reflect service account membership in the group when SA is created with a group" and users_test.go has "should reflect user membership in the group when user is created with a group". So the bidirectional relationship is tested from both sides, but inconsistently organized — the SA/user files test the effect on the group, while the groups file tests the effect on the SA/user. Picking one consistent direction and deduplicating would clean this up.

serviceaccounts_test.go mixed concerns in update test
"should update the service account name" (line ~180) also asserts AccessToken is nil — a separate invariant about token visibility on updates. That’s a distinct behavior worth its own It block (e.g., "should not return access token in update response").

Pre-existing: Project Discovery in organizations_test.go
The Describe("Project Discovery") block lives in organizations_test.go while projects_test.go exists for project CRUD. However, this block is pre-existing — the diff only changes one error type (ErrUnexpectedStatusCode → ErrResourceNotFound) — so not something to address in this PR.

JakeNScale (approved):

peterj-ns:
Addressed the structural feedback from the previous review:

**acl_test.go**
- Merged the two ACL project tests into one. The original "project appears in ACL" test was a strict subset of the deletion lifecycle test — `WaitForProjectInACL` already asserts inclusion as a prerequisite. The remaining test is now a pure deletion lifecycle: create → provision → wait to appear (silent prerequisite) → delete → assert gone.

**groups_test.go**
- Removed cross-resource membership assertions (`Group.Spec.UserIDs`, `Group.Spec.ServiceAccountIDs`). These are owned by `users_test.go` and `serviceaccounts_test.go`, which already test both directions of the relationship from the resource that holds `Spec.GroupIDs`.
- Removed misplaced user/SA membership tests from "When updating groups" — those were testing resource creation with group assignment, not group update behaviour.
- `groups_test.go` now only asserts on group-intrinsic fields: metadata, name, roles.

JakeNScale (approved):

PR #443, opened 2026-04-13
feat: add group subjects migration manager

alansy-nscale:
## Description

This PR introduces a standalone one-off migration tool (`unikorn-group-subjects-migration-manager`) to repair `Subject` data on existing `Group` resources across all organizations. It corrects three classes of data inconsistency: missing `UserID` fields on existing `Subjects` (backfilled via subject email lookup against `OrganizationUser`/`User` records), missing `Subject` entries for `UserIDs` present in the group's `UserIDs` list, and `Subjects` with an empty `Issuer` field (patched to use the configured identity issuer URL).

The tool was initially implemented as a Helm pre-upgrade hook Job but has been refactored into a manually-run standalone binary, giving operators explicit control over when the migration is applied. It supports a `--dry-run` mode for validating changes without writing any patches, resume behaviour that skips already-successful groups on re-runs (making partial failure recovery safe), configurable worker concurrency, and structured per-group JSON output for audit purposes.


spjmurray:
This doesn't belong in identity, surely it should be its own repo that can be discarded without another PR??

spjmurray:
In fact, this looks like you could just get someone with admin powers to run a script...

alansy-nscale:
It was in [another repository](https://github.com/nscaledev/uni-auth0-migration/blob/main/cmd/group-subject-migration-manager/main.go) and was designed to be run by an admin, but I was told to move it into this repository.

squaremo:
> This doesn't belong in identity

Why not? It's responding to a change made here (#433). In particular, deploying a release with that change will break things unless you've done this migration.

It is not (if this is what you think) to facilitate Auth0 integration.

alansy-nscale:
I understand why this should be in this repository, but at the same time, it would be easier to run it as a script rather than as part of the deployment. It is likely a one-off task, and we would then need to remove it in the next release. Plus, it will likely require some human intervention to resolve data integrity issues.

squaremo:
> I understand why this should be in this repository, but at the same time, it would be easier to run it as a script rather than as part of the deployment.

OK :+1: It's already a command-line tool, so you can remove the Helm hook bit, and just document that the tool needs to be run before deploying the breaking change (from #433).

One suggestion if you do that: give the command-line tool a `dry-run` flag, which just reports what it _would have_ done without doing it, and warns if there are any inconsistencies.

alansy-nscale:
> 1. **Memory footprint**: Loads all Groups, OrganizationUsers, and Users into memory (main.go:315-337). For clusters with 10,000+ users across hundreds of orgs, this could consume significant RAM. No pagination or streaming.

It's an acceptable trade-off for a one-off tool. The cluster data is bounded and comfortably fits in memory on any machine it's likely to be run from.

> 3. **Inconsistent data handling**: When `ErrInconsistentData` is returned (e.g., duplicate user subjects, main.go:379-381), the entire migration fails. This is correct for cache building, but means one bad User record blocks migration of all groups. Consider whether to fail-fast or continue with partial data.

It's intentional. If a record is malformed the cluster is already in an unexpected state, and quietly skipping it risks leaving groups partially migrated and harder to reason about later.

> 4. **Error context loss** (main.go:485): When user lookup fails, error message says `"user not found for subject %q"` but doesn't mention which group triggered this. Same issue in `computeMissingSubjects()` (main.go:516, 526). Add group namespace/name to error context.

The group namespace and name are always captured alongside the error in the results file, so there's no ambiguity when investigating failures.

> 5. **Race condition potential** (main.go:391-402): The dispatch pattern uses a buffered semaphore channel. If context is cancelled mid-run, goroutines may still write to the `completed` channel after the receiver exits, causing a deadlock. The `spawnWorkers()` function doesn't respect context cancellation.

Not a race condition or deadlock as claimed. The dispatch loop always reads exactly `len(groups)` results so there's no way for the receiver to exit early. The real issue was that `spawnWorkers` didn't check for cancellation, and it has been fixed by adding an early return that short-circuits pending groups and records them as failed immediately.

> 6. **Misleading field name** (main.go:537): `GroupSubject.ID` is set to `user.Spec.Subject` (an email), not to a user ID. The field name suggests an identifier, but it's actually the subject claim. Same on line 538 where `Email` is also set to `user.Spec.Subject`. This may be intentional (matching existing API), but is confusing. If this is correct, document why ID and Email are both set to Subject.

Both fields holding the same value is intentional. The system uses email as the OIDC subject claim, so ID and Email naturally end up identical.

---

The other valid major issues have been addressed.

squaremo:
```
$ go run ./cmd/unikorn-group-subjects-migration-manager/ --concurrency=1 --dry-run --file /tmp/results.json --context colima --kubeconfig ~/.kube/config --identity-issuer=https://identity.unikorn.nscale.com
progress: 1/2 groups processed
progress: 2/2 groups processed
mikeb@coco ~/space/unikorn/identity (group-subjects-migration)
$ cat /tmp/results.json
cat: /tmp/results.json: No such file or directory
```

* :+1: progress to stderr
* :+1: ran to completion :-)

* ❓ The missing file was unexpected (I thought it would still write something to the results file, even with --dry-run)
* 🤷  I had to tell it the kubeconfig -- usually tools just look in the usual places
* ❓ Did it do anything? I'm unclear whether it did something and didn't report it, or there was nothing to do

alansy-nscale:
> ```
> $ go run ./cmd/unikorn-group-subjects-migration-manager/ --concurrency=1 --dry-run --file /tmp/results.json --context colima --kubeconfig ~/.kube/config --identity-issuer=https://identity.unikorn.nscale.com
> progress: 1/2 groups processed
> progress: 2/2 groups processed
> mikeb@coco ~/space/unikorn/identity (group-subjects-migration)
> $ cat /tmp/results.json
> cat: /tmp/results.json: No such file or directory
> ```
> 
> * 👍 progress to stderr
> * 👍 ran to completion :-)
> * ❓ The missing file was unexpected (I thought it would still write something to the results file, even with --dry-run)
> * 🤷  I had to tell it the kubeconfig -- usually tools just look in the usual places
> * ❓ Did it do anything? I'm unclear whether it did something and didn't report it, or there was nothing to do

In dry run mode, the file flag is ignored and only failed records are written to stderr.

squaremo:
> In dry run mode, the file flag is ignored and only failed records are written to stderr.

If I were running a migration against a unique live database, I would do something like this:

```
10 run it with --dry-run
20 where it reported making a change, sense check the change it says it would make, so I'm happy it's not going to destroy information -- is the migration itself correct
30 check up on any records that caused problems and fix them or the migration tool
40 goto 10
```

Once there are no problems or unexpected changes, run it for real.
Your taste for adventure may vary of course -- I am seldom accused of a deficit in caution, myself.

alansy-nscale:
I did consider writing to a file, but someone might mistake it for a state file and rerun the migration, which could lead to partially migrated records. On second thought, I might add an extra flag to prevent non-dry-run executions from using a dry-run state file.

squaremo:
> I did consider writing to a file, but someone might mistake it for a state file and rerun the migration, which could lead to partially migrated records

Oh, I didn't click that it held state that was consulted later. Ignoring that flag for a dry run is fine -- but you could print more information out to stderr, to support timid people like me :-)

alansy-nscale:
I've updated the dry run to write a state file, but it's marked with `"dry_run": true" and is rejected by real migrations to prevent accidental reuse.

squaremo (approved):
Looks reasonable and has many, many tests (as befitting something that could be run on live systems!)
I know the plan involves this being tested and its rollback also being tested, so I'm happy for it to be merged.

nscale-peter:
This PR is no longer required. The script described has been successfully run and has been in production for 3 days now. 

There is no advantage of merging this into main despite the successful implementation of this script.

PR #449, opened 2026-04-15
feat(oauth2): add RFC 8693 token exchange grant on /oauth2/v2/token

nscale-peter:
## Summary

Ticket: [Add OpenAPI spec for passport token exchange](https://linear.app/nscale-workspace/issue/ID-186/add-openapi-spec-for-passport-token-exchange-endpoint)

Adds the OpenAPI contract for passport token exchange — Phase 2 of the Auth0 migration (ID-186). Exchange is modelled as an [RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693) grant (`urn:ietf:params:oauth:grant-type:token-exchange`) on the existing `/oauth2/v2/token` endpoint rather than a bespoke path.

A stub handler returns `OAuth2InvalidRequest("token exchange is not yet implemented")` to satisfy the dispatcher contract; the real implementation follows in ID-187.

### Why RFC 8693 on `/token` (not a separate `/exchange`)

Prior art (Keycloak 26.2, Auth0 Custom Token Exchange, Zitadel) puts token exchange on `/token` using `grant_type`. A bespoke endpoint is invisible to [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414) / OIDC `.well-known` discovery — the metadata schema has a singular `token_endpoint` and a `grant_types_supported` array, with no field for alternative endpoints. Folding exchange into `/token` keeps us discoverable for free and lets generic OAuth clients use the flow with zero bespoke code.

RFC 8693 §2.2.1 explicitly accommodates non-standard issued tokens: the passport goes in the `access_token` field, and `issued_token_type` identifies the format. The spec literally says *"the issued token need not be an OAuth access token"*.

### Changes

- **Spec** (`pkg/openapi/server.spec.yaml`):
  - Adds `urn:ietf:params:oauth:grant-type:token-exchange` to the `grantType` enum
  - Extends `tokenRequestOptions` with RFC 8693 fields (`subject_token`, `subject_token_type`, `requested_token_type`, `audience`, `resource`, `scope`) plus UNI extensions (`organizationId`, `projectId`)
  - Extends the `token` response with `issued_token_type`
  - Removes the bespoke `/oauth2/v2/exchange` path and its associated schemas (`exchangeRequestOptions`, `exchangeResult`, request/response bodies)
- **OIDC discovery** (`pkg/handler/handler.go`): advertises the new grant in `grant_types_supported`
- **Dispatcher** (`pkg/oauth2/oauth2.go`): `Token()` now routes the exchange grant to a stub `TokenExchange()`
- **Generated code**: types, router, client, mocks regenerated

### Incidental cleanup

Normalizes a handful of field descriptions that referred to the `"code"` grant type to `"authorization_code"` — matching the actual enum value, the handler dispatch (`pkg/oauth2/oauth2.go:1364`), and [RFC 6749 §4.1.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3). `"code"` in OAuth 2.0 is the `response_type` value used at the authorization endpoint, not a `grant_type` — the old prose was misleading. No wire-contract change.

### Non-changes

- Passport TTL (2 minutes) and JWKS (reuses `/oauth2/v2/jwks`) — unchanged, will apply when ID-187 lands
- Passports remain inside the trust boundary (signed, not encrypted, contains full ACL) — unchanged
- Endpoint continues to be marked `x-hidden` and `x-no-security-requirements`

## Test plan

- [x] `make validate` passes
- [x] `make lint` — Go lint 0 issues (Helm lint failure is pre-existing on main, unrelated)
- [x] `make generate` produces expected types, router methods, and client bindings
- [x] `make test-unit` — all test packages pass (Make exit 1 is a pre-existing Go toolchain version mismatch in coverage tooling)
- [x] Discovery endpoint advertises the new grant type
- [x] Dispatcher routes the grant to the stub handler


pkg/openapi/client.go
```diff
@@ -3016,6 +3045,46 @@ func NewPostOauth2V2AuthorizationRequestWithBody(server string, contentType stri
 	return req, nil
 }
 
+// NewPostOauth2V2ExchangeRequestWithFormdataBody calls the generic PostOauth2V2Exchange builder with application/x-www-form-urlencoded body
+func NewPostOauth2V2ExchangeRequestWithFormdataBody(server string, body PostOauth2V2ExchangeFormdataRequestBody) (*http.Request, error) {
+	var bodyReader io.Reader
+	bodyStr, err := runtime.MarshalForm(body, nil)
```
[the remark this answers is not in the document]
nscale-peter: This code is auto-generated by oapi-codegen (`make generate`) — we don't hand-write or modify `client.go`. Any manual fix here would be overwritten on the next generation run.

The same `MarshalForm` + nullable pointer pattern is used by every existing form-urlencoded endpoint in this codebase (`PostOauth2V2TokenWithFormdataBody`, `PostOauth2V2LoginWithFormdataBody`, etc.) — this isn't new to the exchange endpoint.

If the panic is reproducible it would be an upstream oapi-codegen issue, not something to address in a spec PR.

pkg/openapi/server.spec.yaml
```diff
@@ -1511,22 +1512,25 @@ components:
...
         grant_type:
-          description: Supported grant type.  Must be either "code", "refresh_token" or "client_credentials".
+          description: |-
+            Supported grant type. One of "authorization_code", "refresh_token",
+            "client_credentials", or "urn:ietf:params:oauth:grant-type:token-exchange"
+            (RFC 8693).
           type: string
         code:
-          description: Authorization code. Required with the "code" grant type.
+          description: Authorization code. Required with the "authorization_code" grant type.
```
nscale-peter, line 1521: The grantType enum has always been [authorization_code, client_credentials, refresh_token], and the handler has always dispatched on grant_type=authorization_code (see pkg/oauth2/oauth2.go:1364). Only the prose descriptions used the shorthand "code" — for example:

grant_type:
  description: Supported grant type. Must be either "code", "refresh_token" or "client_credentials".
That was inconsistent with the enum and misleading for integrators: per RFC 6749 §4.1.3, the grant_type identifier is authorization_code. The string code in OAuth2 is the response_type value used at the authorization endpoint — a different parameter entirely. So the old descriptions referenced a token that, read literally, was a different OAuth concept.

Clients were already required to send grant_type=authorization_code; anyone who tried grant_type=code based on the description would have hit OAuth2UnsupportedGrantType. This aligns the documentation with the implementation.

Happy to split this out of the token-exchange PR if reviewers would prefer to keep that change focused.

pkg/handler/handler.go
```diff
@@ -152,6 +152,7 @@ func (h *Handler) GetWellKnownOpenidConfiguration(w http.ResponseWriter, r *http
 			openapi.AuthorizationCode,
 			openapi.ClientCredentials,
 			openapi.RefreshToken,
+			openapi.UrnIetfParamsOauthGrantTypeTokenExchange,
```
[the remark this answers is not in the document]
nscale-peter: This PR is for documenting the openAPI spec changes only - this is by design

pkg/openapi/server.spec.yaml
```diff
@@ -1537,6 +1541,60 @@ components:
...
+          type: string
+          nullable: true
+        scope:
+          description: |-
+            A list of space-delimited, case-sensitive strings that specify the
+            requested scope of the issued token. Optional for the token-exchange
+            grant (RFC 8693 section 2.1).
+          type: string
+          nullable: true
+        organizationId:
```
spjmurray: I would probably suggest changing this to `x_organization_id`, or something similar.  Keeps it well out of the way of any other extensions and follows the snake case convention.  Branding things explicitly is just asking for trouble.

I'm in agreement with Claude that this is the best way of doing this, scopes is wrong, and adding it to the path... well we all know how much of a pain it is to use Entra as a result!!

pkg/openapi/server.spec.yaml
```diff
@@ -1537,6 +1541,60 @@ components:
...
+          description: |-
+            UNI extension. Optional organization context to set in the passport's
```
spjmurray: Speaking of branding... remove this please, we want public facing things to be as white-box as possible.

spjmurray (commented):
Just a couple things to muse over.

@claude, please ignore the fact it's advertised but not implemented.  Just because you have inferred something does not net mean it will be deployed and customer facing.  That's may job, not yours.

spjmurray (commented):
Right now for the contentious bits...

* Sneaking in skills was not part of the remit, but _I_ shalltn't block you
* Now you have introducted org/project ID, but -- unless I'm skimming too quickly -- not described _how_ they will be used.  As this is only 10% of the full change I d not know where you are going with this and have to infer... I can imagine those being fully optional, so provide none, you get the full scoped "dynamic" token back, just as we have now.  Specify just org and it's good for that organization only (if allowed by RBAC), likewise projects, but in this case I'm assuming -- given the current project lookup semantics -- you have to also specify the organization??

spjmurray (approved):
For the record, the fields I was interested need not be used, but I can see applications where you may want to bind a token to a specific org/project to limit the scope of use, so I'm giving this the thumbs up

PR #451, opened 2026-04-16
Fix impersonated service account ACL dispatch

spjmurray:
## Summary

Identity currently propagates an impersonated principal with only actor and scope information. When a system service makes an internal ACL request with impersonation enabled, RBAC assumes the propagated actor is always a user and unconditionally routes the impersonated branch through `processUserAccountACL`. That breaks service-account-rooted flows because the impersonated subject is looked up as a `User` and fails when no `User` object exists.

This change carries the existing auth-claim account type through principal propagation and uses it to dispatch impersonated ACL resolution correctly.

## Changes

- add a `type` field to `principal.Principal` using the existing `openapi.AuthClaimsAcctype` values
- set the propagated principal type in the OpenAPI middleware when generating a principal from bearer-token authentication
- route impersonated system-account ACL resolution by propagated principal type:
  - `user` -> `processUserAccountACL`
  - `service` -> `processServiceAccountACL`
- fail closed for missing, unknown, or `system` impersonated principal types
- update supporting fixtures and tests so propagated principals are typed consistently
- add regression coverage for impersonated service-account ACL resolution and the missing-type failure path

## Validation

- `GOCACHE=/tmp/gocache go test ./pkg/middleware/openapi ./pkg/rbac`
- `GOCACHE=/tmp/gocache go test ./pkg/handler/common ./test/contracts/provider/region`
- `GOCACHE=/tmp/gocache make validate`
- `GOCACHE=/tmp/gocache make lint`

## Not Run

- integration tests

### DIFF
--- a/pkg/handler/common/fixtures/fixtures.go
+++ b/pkg/handler/common/fixtures/fixtures.go
@@ -46,6 +46,7 @@ func HandlerContextFixture(ctx context.Context, flags int) context.Context {
 	}
 
 	p := &principal.Principal{
+		Type:  openapi.User,
 		Actor: PrincipalActor,
 	}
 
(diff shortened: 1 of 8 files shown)


spjmurray:
Critical test blocker, proven to work via integration testing.

squaremo (approved):
- [x] the explanation of the problem makes sense
 - [x] the changes are minimal and look like they solve the problem
 - [x] tests demonstrate that the problem is solved

PR #453, opened 2026-04-16
Add load balancer RBAC entries and quota defaults

mkjpryor:
Adds `region:loadbalancers:v2` permissions to the existing organization- and project-scoped roles in the identity chart values. Adds a default `loadbalancers` quota with display metadata and a default limit of 5. Tests were not run in this workspace.


squaremo (approved):
Looks reasonable; in particular, it lines up with analogous resources.

spjmurray (approved):

PR #456, opened 2026-04-16
Add SSH CA read access to compute-service

spjmurray:
The compute service checks SSH certificate authority references while impersonating the user. That RBAC decision is the boolean intersection of the service role and the impersonated user permissions, so compute-service also needs read access to region:sshcertificateauthorities:v2 or the check fails even when the user is allowed to reference the project-scoped CA.


spjmurray:
Confirmed as working via integration testing.

PR #461, opened 2026-04-22
feat(passport): add mTLS impersonation support to token exchange

nscale-peter:
## Summary

Implements ID-189 (step 4 of 7 in the Phase 2 Passport Token Exchange rollout). Services can now present a client cert plus `X-Impersonate: true` and an `X-Principal` header on `/oauth2/v2/exchange` and receive a passport minted on the impersonated user's behalf.

- Passport `acctype` is `user` with `actor` = impersonated subject so audit surfaces the real principal, not the service.
- Passport ACL is the intersection of the user's and the service's ACLs (confused-deputy prevention), computed via the existing `rbac.GetACL` system-account branch.
- Autonomous mTLS (cert without `X-Impersonate`) is refused with `invalid_request` so service-to-service traffic can't silently upgrade to system creds.
- The `impersonated=true` flag and `serviceIdentity` are emitted on the `passport_issued` log.
- Shared `principal.ExtractFromRequest` helper replaces the duplicate inline parser in the openapi middleware; now covered by its own unit test suite.

## Stack

Depends on #450 (`nscale-peter/id-187-passport`). This PR targets that branch so the full stack merges in order:
- #449 → `petermcarthur/id-186` (openapi spec)
- #450 → `nscale-peter/id-187-passport` (exchange endpoint)
- This PR → mTLS impersonation on top

## Test plan

- [x] `make touch license validate lint` — clean
- [x] `go test ./pkg/oauth2/... ./pkg/principal/... ./pkg/middleware/openapi/...` — passes
- [x] New unit tests in `pkg/oauth2/passport_mtls_test.go` (8 scenarios: happy path, missing/false `X-Impersonate`, malformed principal, empty actor, out-of-scope org, ACL intersection, missing cert)
- [x] New unit tests in `pkg/principal/extract_test.go` (6 scenarios)
- [x] BDD integration specs added to `test/api/suites/passport_test.go` under `When a service impersonates a user via mTLS` (happy path + 4 negative cases)
- [ ] `make test-api` against a KinD cluster after CI provisions the `ci-impersonator` fixture

### DIFF
--- a/hack/ci/fixtures/main.go
+++ b/hack/ci/fixtures/main.go
@@ -248,6 +248,60 @@ func createProject(ctx context.Context, ac *openapi.ClientWithResponses, orgID, 
 	return id
 }
 
+// createUser creates a user in the given groups and returns its ID.
+func createUser(ctx context.Context, ac *openapi.ClientWithResponses, orgID, subject string, groupIDs []string) string {
+	logf("Creating user %q...", subject)
+
+	resp, err := ac.PostApiV1OrganizationsOrganizationIDUsersWithResponse(ctx, orgID, openapi.UserWrite{
+		Spec: openapi.UserSpec{
+			Subject:  subject,
+			State:    openapi.Active,
+			GroupIDs: groupIDs,
+		},
+	})
+	if err != nil {
+		fatalf("failed to create user %q: %v", subject, err)
+	}
+
+	if resp.JSON201 == nil {
+		fatalf("create user %q returned %s", subject, resp.Status())
+	}
+
+	id := resp.JSON201.Metadata.Id
+	logf("  user %q ID: %s", subject, id)
+
+	return id
+}
+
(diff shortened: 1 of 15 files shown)


nscale-peter:
Closing — passport scope reduced to bearer-token entry path only.

Per discussion with @nscale-peter: main's mTLS + X-Impersonate + X-Principal flow never carries a bearer token, so there is nothing for the passport to replace on this path. Identity's existing intersection logic in `getSystemAccountACL` runs against `(service-CN, X-Principal-actor)` on each ACL fetch and preserves confused-deputy guarantees without modification. Bringing passport into the impersonation flow added complexity (the privilege-escalation problem from removing the embedded ACL) without solving any Phase 2 problem.

PRD §9 will be updated to specify that mTLS flows are unchanged. Downstream passport middleware will skip passport logic when no Authorization header is present and fall through to the existing mTLS path.

ID-189 closed as cancelled.

PR #463, opened 2026-04-23
fix(openapi): separate direct and impersonated ACL cache entries

spjmurray:
## Summary
- separate direct/attributed ACL cache entries from impersonated ACL cache entries
- retain organization scope in the cache key for scoped ACL responses
- reject malformed impersonation requests at the middleware boundary and add cache regression coverage

## Problem
The ACL middleware cache key did not fully encode how an ACL had been resolved. Direct bearer-token requests and impersonated service-to-service requests could collide when they referred to the same actor and organization, even though impersonated ACLs are intersections of service and actor permissions rather than the actor's full ACL.

In practice that meant an impersonated ACL could poison the cache for a later direct request and under-authorize it.

## Fix
The cache key now distinguishes authorization mode:
- `direct|<authenticated-subject>|<org-or-_global>` for direct bearer-token and attributed non-impersonated service calls
- `impersonated|<calling-service-subject>|<actor>|<org-or-_global>` for impersonated service calls

The requested organization remains part of the key because scoped `/api/v1/organizations/{id}/acl` responses still depend on org scope.

This also tightens the impersonation model: if `X-Impersonate` is set but the propagated principal has no actor, the middleware now rejects the request with `400` instead of silently degrading to a direct path.

## Tests
Added and updated middleware tests covering:
- direct cache hits
- impersonated cache hits
- direct vs impersonated separation
- different calling services not sharing impersonated cache entries for the same actor
- malformed impersonation being rejected at the boundary

## Verification
- `make touch`
- `make license`
- `make validate`
- `make generate`
- `make lint`
- `GOCACHE=/tmp/gocache go test ./pkg/middleware/openapi ./pkg/rbac`
- `GOCACHE=/tmp/gocache make test-unit`
- local Claude review against `main`

### DIFF
--- a/pkg/middleware/openapi/cachekey_test.go
+++ b/pkg/middleware/openapi/cachekey_test.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2026 Nscale.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//nolint:testpackage // We intentionally exercise the unexported cache-key helper directly.
+package openapi
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/unikorn-cloud/identity/pkg/middleware/authorization"
+	identityapi "github.com/unikorn-cloud/identity/pkg/openapi"
+	"github.com/unikorn-cloud/identity/pkg/principal"
+)
+
+func TestACLCacheKey(t *testing.T) {
+	t.Parallel()
+
+	// directInfo models a direct bearer-token call. The same cache key shape is
+	// also used for attributed service-to-service calls where the principal
+	// header is audit-only and RBAC still resolves against the authenticated
+	// caller.
+	directInfo := &authorization.Info{
+		Userinfo: &identityapi.Userinfo{
+			Sub: "user-1",
+		},
+	}
+
+	serviceInfo := &authorization.Info{
+		Userinfo: &identityapi.Userinfo{
+			Sub: "compute-service",
+		},
+		SystemAccount: true,
+	}
+
+	t.Run("DirectIncludesOrganizationScope", func(t *testing.T) {
+		t.Parallel()
+
+		global, err := aclCacheKey(t.Context(), directInfo, "")
+		require.NoError(t, err)
+
+		scoped, err := aclCacheKey(t.Context(), directInfo, "org-1")
+		require.NoError(t, err)
+
+		require.Equal(t, "direct|user-1|_global", global)
+		require.Equal(t, "direct|user-1|org-1", scoped)
+		require.NotEqual(t, global, scoped)
+	})
+
+	t.Run("AttributedCallUsesDirectKeyShape", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := principal.NewContext(t.Context(), &principal.Principal{
+			Actor: "someone-else",
+		})
+
+		key, err := aclCacheKey(ctx, serviceInfo, "org-1")
+		require.NoError(t, err)
+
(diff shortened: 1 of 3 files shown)


spjmurray:
Long story short what we have hypothesized has happened.  User will have asked the compute service to list compatible images and flavors impersonating the user (thus correctly scoped).  The ACL in region is effectively the intersection between what the compute service can do and what the user can do.  This was cached.  Later on the user directly wanted to delete the image but unfortunately the cache key aliased with the earlier effective one, and as the compute service cannot delete images, neither can the user and pow, 403.

The fix is simply to ensure cache keys fully handle all possible ACL scopes (that matter) to RBAC.

spjmurray:
Will cobble together a quick RC4 point release to verify it works.

spjmurray:
Working: https://github.com/nscaledev/uni-compute/actions/runs/24854134394
Not working: https://github.com/nscaledev/uni-compute/actions/runs/24842733212

pkg/middleware/openapi/openapi_test.go
```diff
@@ -478,6 +529,160 @@ func TestServiceToServiceAuthenticationSuccess(t *testing.T) {
...
+	addCertificateHeader(t, r, true)
+	addPrincipalHeaderWithoutActor(t, r)
+	r.Header.Set(principal.ImpersonateHeader, "true")
+
+	m.ServeHTTP(w, r)
+
+	validateError(t, w, errors.IsBadRequest)
+}
+
+func TestServiceToServiceDirectAndImpersonatedACLsDoNotShareCacheEntries(t *testing.T) {
```
squaremo, line 557: It is not that obvious how the test body verifies the property under test. Is it that GetACL is called twice, meaning it hasn't got a result from the cache on the second call?
A more obvious way to demonstrate it might be to arrange for a different authorisation outcome when impersonating.

pkg/middleware/openapi/openapi_test.go
```diff
@@ -478,6 +529,160 @@ func TestServiceToServiceAuthenticationSuccess(t *testing.T) {
...
+	require.NoError(t, err)
+	addCertificateHeader(t, r2, true)
+	addPrincipalHeader(t, r2)
+
+	w2 := httptest.NewRecorder()
+	m.ServeHTTP(w2, r2)
+	require.Equal(t, http.StatusOK, w2.Result().StatusCode)
+}
+
+func TestServiceToServiceImpersonatedACLsReuseCacheEntries(t *testing.T) {
```
squaremo, line 620: On the surface of it, it looks like these tests could share some setup/teardown and be more like a table test -- especially these "reuse" ones. Maybe the specialisation of each case ends up being more complicated than just spelling out the whole thing?

pkg/middleware/openapi/cachekey_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	serviceInfo := &authorization.Info{
+		Userinfo: &identityapi.Userinfo{
+			Sub: "compute-service",
+		},
+		SystemAccount: true,
+	}
+
+	t.Run("DirectIncludesOrganizationScope", func(t *testing.T) {
```
squaremo, line 50: These look like "change detection" tests -- that is, an equally valid implementation of cache keys (like using a different separator) would fail them, but not represent a real problem in the system.

I think you want to check _properties_ here rather than return values. That is: different orgs -> different cache keys, impersonation v not -> different cache keys, that sort of thing.

pkg/middleware/openapi/openapi.go
```diff
@@ -149,25 +148,55 @@ func hasHTTPAuthorization(r *http.Request) bool {
 	return r.Header.Get("Authorization") != ""
 }
 
-// aclCacheKey returns the key to use when caching an ACL. For impersonated calls
-// the key includes the end-user actor and their organization IDs so that each
-// user gets their own cache entry and different org claim sets produce distinct
-// entries. Organization scope must also be part of the key, otherwise unscoped
-// ACLs can be incorrectly reused for /organizations/{id}/acl and other scoped routes.
-func aclCacheKey(ctx context.Context, info *authorization.Info, organizationID string) string {
+// aclCacheKey returns the key to use when caching an ACL.
```
squaremo, line 151: This is an exemplary explanation :clap:

squaremo (approved):
:+1:, though please read my comments and consider whether anything needs following up.

(This could have been smaller changes in theory, but it doesn't make much difference to legibility in this case.)

spjmurray:
Gracias, ticket raised for tracking.

PR #469, opened 2026-04-27
feat(passport): passport-first authorization with local verification and token exchange

alansy-nscale:
## Description

This branch builds on `ID-188` and moves passport middleware from verification-only behaviour to a passport-first authorization flow.

For non-passport bearer tokens, middleware now exchanges the source token for a passport, verifies the minted passport locally, and derives authorization context from verified passport claims. For incoming passport tokens, middleware verifies locally and proceeds without exchange.

Authorization remains fail-closed for invalid or expired credentials. Degraded behaviour is limited to token-exchange unavailability (transport/upstream failures), where middleware delegates to the configured fallback authorizer. The fallback is now explicitly injected at construction time, allowing embedding services to choose behaviour appropriate for their runtime (for example, identity-service wiring).

To support downstream propagation safely, authorization context now carries both source token and passport when exchange occurs. Outbound injection continues to propagate `X-Principal` and additionally sets `Authorization: Bearer <passport>` when passport context is present.

For remote ACL lookups, passport-only contexts do **not** forward passport JWTs as bearer credentials. ACL evaluation instead uses delegated-principal impersonation semantics, so RBAC behaviour stays aligned with the platform model.

### New `pkg/middleware/openapi/passport` package

- `verifier.go` — local passport JWT verification: ES512 signature check via JWKS `kid`, standard JWT claim validation, defence-in-depth `typ` re-check, and `isPassport` payload-only routing hint that lets non-passport tokens fall through to exchange.
- `keysource.go` / `keysource_cached_http.go` / `keysource_local.go` — `KeySource` interface with two implementations: cached HTTP JWKS client (with refresh and stale-tolerance behaviour) and an in-process source for the identity service itself.
- `token_exchange.go` / `token_exchange_http.go` / `token_exchange_local.go` — `TokenExchange` interface with HTTP (RFC 8693 form POST against `/oauth2/v2/token`) and in-process implementations. Exchange errors are classified into sentinels (`ErrTokenExchangeUnauthorized`, `ErrTokenExchangeUnavailable`, `ErrTokenExchangeFailed`, `ErrTokenExchangeInvalidResponse`, `ErrTokenExchangeMissingAccessToken`) so the authorizer can route between fail-closed and fallback paths.
- `authorizer.go` — implements `openapi.Authorizer`: verify-then-fall-through-to-exchange, fail-closed on credential errors, fallback to the injected `uni` authorizer only on `ErrTokenExchangeUnavailable`. `GetACL` is delegated to the fallback. Prometheus metrics for verification, authorizer dispatch, exchange outcomes, and exchange latency.
- `urls.go` — `JWKSURL` and `TokenExchangeURL` helpers.
- Comprehensive unit tests across all components.

### Authorization context (`pkg/middleware/authorization/authinfo.go`)

- `Info` gains a `Passport` field so handlers can carry both the original source token and the exchanged passport.

### Principal injector (`pkg/principal/injector.go`)

- When the active authorization context carries a passport, `Injector` now also sets `Authorization: Bearer <passport>` on outbound internal calls in addition to `X-Principal`.
- Adds `PrincipalOnlyInjector` for call sites that must propagate principal/impersonation but *not* the bearer credential.

### Remote authorizer (`pkg/middleware/openapi/remote/authorizer.go`)

- For passport-only ACL lookups, switches from bearer propagation to delegated-principal impersonation semantics: the call uses service identity (mTLS) plus `X-Principal` and `X-Impersonate: true`, so the callee resolves RBAC against the delegated end-user principal rather than trusting passport propagation between internal services.

### Bug fix

- HTTP token-exchange form fields now use `x_organization_id` / `x_project_id` to match the server-side `TokenRequestOptions` schema. The previous `organizationId` / `projectId` names were silently dropped by the form parser, which would have caused passports to be minted without the route's intended scope narrowing.


alansy-nscale:
>  
    #### 5. **PR Description Alignment Issue**
  
    **Discrepancy**: The PR description states:
    
    > "For remote ACL lookups, passport-only contexts no longer forward passports as bearer credentials"
    
    **But the code** (pkg/middleware/openapi/remote/authorizer.go:282) shows:
    ```go
    } else if info.Passport != "" {
        requestCtx = principal.NewImpersonateContext(requestCtx)
    }
    ```
    
    This **enables impersonation**, it doesn't "no longer forward passports." The passport IS still present in the context (propagated via `Injector` in pkg/principal/injector.go:53-55):
    
    ```go
    if authInfo, err := authorization.FromContext(ctx); err == nil && authInfo.Passport != "" {
        r.Header.Set("Authorization", "bearer "+authInfo.Passport)
    }
    ```
    
    **Clarification needed**: Is the PR description misleading, or is there a subtlety I'm missing about when `PrincipalOnlyInjector` vs `Injector` is used?
    
The remote authorizer now uses `PrincipalOnlyInjector`, which never injects the passport into the `Authorization` header. In passport-only contexts it sets delegated impersonation (`X-Principal` + `X-Impersonate`) instead. `Authorization` is only set when a source token (`info.Token`) is present.

alansy-nscale:
> Security equivalence: Can you document the security properties of the remote authorizer fallback path vs. local passport verification? Under what conditions would an operator want to disable fallback entirely?

Fallback on `ErrExchangeUnavailable` keeps the legacy path, which still requires valid bearer token authentication and ACL authorization. It is an availability degradation path, [not an authorization bypass](https://github.com/nscaledev/uni-identity/blob/fc7fc6111c3a341f7a822c3986eceac44e8fb2aa/pkg/middleware/openapi/passport/authorizer.go#L251). If the exchange is unavailable, authentication still goes through remote validation and ACL checks. An attacker forcing degraded mode would increase dependency and latency, but would not gain any additional privileges without valid credentials.

alansy-nscale:
> ACL impersonation semantics: Can you explain (ideally in inline comments) why passport-only authorization contexts trigger impersonation mode rather than direct bearer forwarding?

`GetACL` is an internal service-to-service call, so in passport-only cases we intentionally use the platform's delegated-principal model instead of another bearer-token hop. By that stage we already have verified user identity from the passport, so the service authenticates with mTLS and carries user context via `X-Principal` + `X-Impersonate`. Forwarding the passport as `Authorization: Bearer ...` on the ACL call would create a second internal bearer path and blur the service+delegation boundary. We still use bearer when `info.Token` is present for backward compatibility, but passport-only is deliberately handled through impersonation semantics.

pkg/middleware/openapi/passport/keysource_cached_http.go
```diff
@@ -0,0 +1,165 @@
...
+		Name: "identity_passport_jwks_cache_refresh_total",
+		Help: "Total number of JWKS cache refresh attempts, labeled by trigger and result.",
+	},
+	[]string{"trigger", "result"},
+)
+
+// CachedHTTPKeySource fetches and caches the identity service's public JWKS.
+// It refreshes on TTL expiry or on a key-ID miss. Zero network calls on the
+// hot path when a valid cached key is present.
+type CachedHTTPKeySource struct {
```
spjmurray, line 46: This looks a lot like a generic timeout cache, which could have a user-defined callback function... sounds a lot like what already exists in core, but with some tweaks.  You could imagine:

```go
core/pkg/cache.NewTimeoutWithDiscriminatorAndRefreshFunc[*jose.JSONWebKeySet]("passport_cache", time.Minute, myCallback)
```

Then everyone can use it, everyone gets metrics (using tags or something in this case?), you fix once and you fix everywhere.

spjmurray (commented):
While I consider the meat of this whole thing, there's a little observation I've made, that may improve the lives of everyone

pkg/middleware/openapi/passport/keysource_cached_http.go
alansy-nscale: Should we just keep it as it is for now? This [commit](https://github.com/nscaledev/uni-identity/commit/5f2867e94dbf41589250fc404e71388344165a1f#diff-2d51276879c1e7f8d23bf0954171dd5c6685a209e778de8cdcc07684c158a05cR52) uses this specific `CachedHTTPKeySource`, and it does a bit more than a timeout cache. It refreshes on TTL expiry and on kid misses, and it also has singleflight built in so we do not make concurrent network calls to the JWKS endpoint.

nscale-peter:
This is superseded by https://github.com/nscaledev/uni-identity/pull/490

PR #476, opened 2026-05-06
feat: migrate identity auth wiring to passport

alansy-nscale:
## Description

This PR builds on `ID-212` and updates the `uni-identity` API auth wiring to use passport middleware.


nscale-peter:
Given the revised scope Phase 2 should not wire `uni-identity`'s own API middleware through the passport authoriser. It also removes the downstream local-verification model that this PR builds on. The agreed Phase 2 path is to keep existing mTLS + `X-Principal` service-to-service behaviour unchanged and replace direct userinfo validation in middleware with exchange-backed passport data.

This PR was a good implementation of the initial scope however that has since moved on

PR #482, opened 2026-05-12
fix: deduplicate GroupSpec identity lists before persistence

alansy-nscale:
## Description

This PR makes group writes deduplicate GroupSpec list fields before persistence, so duplicate `roleIDs`, `serviceAccountIDs`, `subjects`, and derived `userIDs` are not stored in Kubernetes state. The write path is now normalized through focused helpers, including extracted role validation logic, to keep behavior explicit and maintainable.

It also adds tests for deduplication and persistence behavior, and updates the groups package docs to record the new invariant that these lists are normalized to first-occurrence unique values. No API schema changes are introduced.


nscale-peter (approved):

spjmurray (approved):
Sounds like a good plan!

PR #483, opened 2026-05-12
test(api): cover group API with subject userID mutations

sravanmedarapu-work:
Focused split from #462.

Includes:
- add a typed group payload builder helper for subjects
- add integration coverage for creating groups with external subjects and legacy userIDs
- assert legacy userID creates populate the exact userID and resolved subject fields
- cover invalid userID rejection, subject/userID conflict rejection, GET subject fields, and PUT add/remove subject mutations

Validation:
- GOCACHE=$PWD/.codex-gocache go test ./test/api/...

PR #484, opened 2026-05-12
test(api): cover audit, userinfo, and RBAC access

sravanmedarapu-work:
Focused split from #462 for audit, userinfo, and account-type RBAC integration coverage.

Scope:
- Extends fixture generation only where needed for this PR-owned coverage: audit token, real user token, user/service-account IDs, and non-member organization ID.
- Adds typed API client/endpoints support for userinfo and the RBAC denial paths used by the tests.
- Adds userinfo coverage for fixture, user, and service-account tokens, including stable identity, account type, and organization-scope claims.
- Adds audit-token read coverage for console-visible identity resources and exact 403 coverage for write operations.
- Tightens RBAC matrix assertions so user-denied paths assert 403 instead of any error, and admin/audit reads assert response body fields.
- Extends service-account ACL coverage to assert the home org is present and the non-member org is absent.

Out of scope:
- No product behavior changes.
- No broad fixture foundation split or unrelated fixture refactor.
- No failure response body logging; that remains a separate low-priority PR.
- No group subject/userID runtime fix; that was handled in #483.

Validation:
- git diff --check
- go test ./hack/ci/fixtures/...
- go test -run "^$" -tags=integration ./test/api/...
- golangci-lint run ./hack/ci/fixtures/... ./test/api/...
- make touch
- make validate
- make generate
- GOTOOLCHAIN=go1.25.8 make test-unit

### DIFF
--- a/hack/ci/fixtures/main.go
+++ b/hack/ci/fixtures/main.go
@@ -71,12 +71,7 @@ func logf(format string, args ...interface{}) {
 	fmt.Fprintf(os.Stderr, "==> "+format+"\n", args...)
 }
 
-func newKubernetesClient() client.Client {
-	cfg, err := config.GetConfig()
-	if err != nil {
-		fatalf("failed to get kubeconfig: %v", err)
-	}
-
+func newClientScheme() *runtime.Scheme {
 	scheme := runtime.NewScheme()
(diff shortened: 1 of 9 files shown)


sravanmedarapu-work:
@claude review.

alansy-nscale (approved):

PR #487, opened 2026-05-15
Reuse existing organization memberships

alansy-nscale:
## Description

This PR makes user creation idempotent for organization membership. If a global user already exists for a subject, the users client now reuses the matching OrganizationUser instead of creating a duplicate membership. This keeps repeated create requests for the same subject from producing multiple organization user records, while preserving the existing group reconciliation behavior. A unit test covers the duplicate-subject path and verifies that only one global user and one organization user are stored.


pkg/handler/users/client.go
```diff
@@ -685,6 +686,51 @@ func (c *Client) getOrCreateGlobalUser(ctx context.Context, request *openapi.Use
...
+		constants.OrganizationLabel: organization.ID,
+		constants.UserLabel:         globalUserID,
+	})
+
+	result := &unikornv1.OrganizationUserList{}
+	if err := c.client.List(ctx, result, &client.ListOptions{Namespace: organization.Namespace, LabelSelector: selector}); err != nil {
+		return nil, fmt.Errorf("%w: failed to list organization users", err)
+	}
+
+	switch len(result.Items) {
```
spjmurray, line 700: As an aside we can now create deterministic resource IDs (based on some namespace UUID and some invariant data) so this pattern is now defunct!  But that would need a migration of some variety and I'm assuming this is all "legacy" anyway.

spjmurray (approved):
Still racy in the TOCTOU sense, but it makes problems less likely, as I say there is now an atomic way of doing this, but it's waaaaaaaaay out of scope!  Approve.

PR #490, opened 2026-05-21
Implement passport token exchange phase

nscale-peter:
## Context
This PR implements the Phase 2 requirement from [PRD: Passport Token Exchange (Phase 2)](https://www.notion.so/nscalecloud/PRD-Passport-Token-Exchange-Phase-2-335caf6bfadc803c8308f251eab3365f): centralise bearer-token validation at the identity token-exchange boundary while keeping the existing mTLS plus `X-Principal` service-to-service model unchanged.

## What Changed
- Replaces the remote middleware bearer-token identity path from direct OIDC/userinfo lookup to RFC 8693 token exchange against identity's existing `POST /oauth2/v2/token` endpoint.
- Adds an HTTP token-exchange client and fail-closed exchange error handling for rejected, unavailable, or malformed exchange responses.
- Decodes the returned UNI passport payload into the existing `authorization.Info.Userinfo` shape so handler-facing authorization and principal code remains unchanged.
- Reuses the existing remote token cache location with typed `uni-core/pkg/util/cache`, keyed by source token plus organization/project route scope, and caches passport-derived identity data rather than userinfo payloads.
- Preserves the existing remote ACL/RBAC flow; ACLs are not embedded in passports.
- Preserves internal service-to-service behavior: no passport propagation, no downstream JWKS verification, and no outbound `Authorization: Bearer <passport>` injection.

## Requirement Mapping
- **Centralised token-family dispatch:** downstream services now ask identity to exchange the source token instead of learning source-token semantics locally.
- **Passport payload as the internal identity shape:** remote middleware consumes `sub`, `acctype`, `source`, `email`, `org_ids`, and optional org/project context from the exchanged passport.
- **Cache behavior:** cache entries are derived from passport expiry with a 10s fudge factor and scoped by route context to avoid reusing unscoped identity for scoped requests.
- **Lifetime safety:** identity now caps passport expiry by the source token expiry, so a cached passport-derived identity cannot outlive the bearer token that produced it.
- **Phase 2 boundaries:** Auth0 live-token acceptance, passport propagation, downstream JWKS verification, new cache infrastructure, and service-to-service protocol changes remain out of scope.

## Validation
- `make lint`
- `go test ./pkg/oauth2 ./pkg/middleware/openapi/remote`

## Notes
The integration passport expectations were updated for the Phase 2 60s passport TTL. Package documentation was updated to describe the exchange-backed remote middleware path and its Phase 2 trust boundaries.


pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -20,40 +20,54 @@ package authorizer
...
+	// when deriving cache TTLs from passport expiry.
+	cacheTTLFudge = 10 * time.Second
+)
+
+// tokenCacheKey includes route scope because passports include requested
+// organization/project context in their claims.
+type tokenCacheKey struct {
+	token          string
+	organizationID string
+	projectID      string
```
spjmurray: The remote authorizer's token cache has two related smells. First, the cache value type is *identityapi.Userinfo even though the data comes from a passport — passportToUserinfo projects passport claims onto the old handler-facing shape, discarding the richer passport fields (OrgID, ProjectID, Source, Actor) in the process; the right end state is to cache the passport claims directly and have handlers consume them. Second, the cache key is a composite of {token, organizationID, projectID} where the org/project values are lifted from URL path parameters — a V1-only convention that breaks silently for V2 APIs which carry no such path params — but the composite key is also pointless regardless, because passportToUserinfo never reads OrgID or ProjectID, making the cached value identical for any org/project combination given the same token; the key should be the token alone.

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -162,94 +169,149 @@ func (a *Authorizer) authorizeOAuth2(r *http.Request) (*authorization.Info, erro
...
+	}
 
-	if err := json.Unmarshal(body, claims); err != nil {
-		return nil, err
+	return claims, nil
+}
+
+// passportToUserinfo projects passport claims onto the existing handler-facing
+// identity shape.
+func passportToUserinfo(claims *oauth2.PassportClaims) *identityapi.Userinfo {
```
spjmurray, line 257: Would a receiver on the PassportClaims type not be cleaner?

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -162,94 +204,186 @@
...
+// here for the same reason as the passport — trust is at the channel layer,
+// and the source token's identity was already validated by the exchange call.
+func sourceTokenExpiry(token string) (time.Time, bool) {
+	parsed, err := jwt.ParseSigned(token, []jose.SignatureAlgorithm{jose.ES512})
+	if err != nil {
+		return time.Time{}, false
+	}
+
+	var claims jwt.Claims
+	if err := parsed.UnsafeClaimsWithoutVerification(&claims); err != nil {
```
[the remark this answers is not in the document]
spjmurray: UnsafeClaimsWithoutVerification is safe here. The passport is not a user-controlled input — it is the response body from the authorizer's own outbound call to identity's token exchange endpoint. The client never touches it. Verifying the signature against identity's JWKS would not add defence-in-depth: both the passport and the JWKS originate from the same identity endpoint over the same TLS connection, so if that connection were compromised an attacker could forge both simultaneously. The real trust boundary is TLS to the identity service, which is already established by the HTTP client setup. ParseSigned with the ES512 algorithm allowlist is sufficient to reject structurally malformed tokens before claims are extracted.

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -162,94 +169,149 @@ func (a *Authorizer) authorizeOAuth2(r *http.Request) (*authorization.Info, erro
...
+		return nil, errors.AccessDenied(r, "token exchange failed").WithValues("error", err.Error())
 	}
 
-	// Do the call manually here to allow us to extract the correct error code and
-	// headers, returned by identity.
-	request, err := http.NewRequestWithContext(r.Context(), http.MethodGet, provider.UserInfoEndpoint(), nil)
+	claims, err := decodePassportClaims(passport)
 	if err != nil {
-		return nil, fmt.Errorf("%w: failed to create userinfo request", err)
+		return nil, errors.AccessDenied(r, "passport could not be decoded").WithValues("error", err.Error())
```
spjmurray: Don't wrap this error, it's internal only and worthy of a 500 e.g. use fmt.Errorf

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -162,94 +169,149 @@ func (a *Authorizer) authorizeOAuth2(r *http.Request) (*authorization.Info, erro
...
-	// and also return some information about the user that we can use for audit logging.
-	provider, err := oidc.NewProvider(ctx, a.options.Host())
+	passport, err := a.exchange.Exchange(ctx, rawToken, &exchangeOptions)
 	if err != nil {
-		return nil, fmt.Errorf("%w: oidc service discovery failed", err)
+		if goerrors.Is(err, ErrTokenExchangeUnauthorized) {
+			return nil, errors.AccessDenied(r, "token is invalid or has expired")
+		}
+
+		return nil, errors.AccessDenied(r, "token exchange failed").WithValues("error", err.Error())
```
spjmurray: There is a `WithError` receiver you should be using surely?

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -65,21 +79,21 @@ func NewAuthorizer(client client.Client, options *identityclient.Options, client
 		return nil, err
 	}
 
+	tokenCache := cache.NewLRUExpireCache[tokenCacheKey, *identityapi.Userinfo](tokenCacheSize)
+	tokenCache.ZeroCopy()
```
spjmurray: Probably best not to use this, just for safety.  It's small enough that DeepCopy wont hurt much.  For ACLs it makes a difference!  But that said, you aren't getting your 👍🏻 until I've run some benchmarks 👿 

Does remind me, we should probably enforce some type of interface on cache data that does require a performant DeepCopy, not just some reflect based cludge!

pkg/middleware/openapi/remote/authorizer.go
```diff
@@ -162,94 +169,149 @@ func (a *Authorizer) authorizeOAuth2(r *http.Request) (*authorization.Info, erro
...
 
-		var err coreapi.Error
+	if claims.Type != oauth2.PassportType {
+		return nil, fmt.Errorf("%w: unexpected token type %q", ErrPassportInvalid, claims.Type)
+	}
 
-		if err := json.Unmarshal(body, &err); err != nil {
-			return nil, fmt.Errorf("%w: failed to unmarshal userinfo error response", err)
-		}
+	if claims.Subject == "" || claims.Expiry == nil {
```
spjmurray, line 225: You're checking `exp` but not `nbf`

spjmurray (commented):
First pass complete!

spjmurray (approved):

PR #494, opened 2026-05-28
fix: map passport-exchange scope refusals to invalid_scope

nscale-peter:
## Summary

After the passport exchange phase landed (#490, shipped in v1.17.0), scope refusals during token exchange were surfacing as `401 access_denied` — indistinguishable from a rejected subject token. Caller-side refresh-loop logic cannot tell "your token is bad, reauthenticate" apart from "your token is fine, you don't have access to that scope" when both arrive as 401.

This change:

- Identity returns `400 invalid_scope` (RFC 6749 §5.2) when a valid subject token requests an organization or project the principal is not a member of.
- The remote middleware (`pkg/middleware/openapi/remote`) classifies the token-endpoint response and projects `invalid_scope` to `403 forbidden` at the downstream API edge, while keeping bad-token failures as `401 access_denied`.
- Adds a new `ErrTokenExchangeForbidden` sentinel and a unit test that pins the 401-vs-403 mapping in the authorizer, plus token-exchange HTTP coverage for the `invalid_scope` / malformed-body cases.
- Documents the authn-vs-authz split in `pkg/oauth2/README.md` and `pkg/middleware/openapi/README.md`.

## Why this was caught

The uni-region dependency bump https://github.com/nscaledev/uni-region/pull/531 (bumping `unikorn-cloud/identity` to v1.17.0) fails its Integration suite on:

> `Region Discovery > When listing regions > Given invalid parameters > should reject requests with invalid organization ID format`
> ([regions_test.go:78](https://github.com/nscaledev/uni-region/blob/deps/bump-identity-v1.17.0/test/api/suites/regions_test.go#L78))

The test asserts the error body contains `"forbidden"`, but with v1.17.0 it received:

```
{"error":"access_denied","error_description":"token is invalid or has expired"}
```

That is the regression introduced by the passport phase: a previously-`forbidden` (403) scope refusal became `access_denied` (401). This PR aligns identity back with the behaviour uni-region (and other downstream callers) were already relying on, and locks in the distinction at the OAuth2 layer per RFC 6749 §5.2 so it cannot silently regress again.

## Test plan

- [x] `make touch && make license && make validate && make lint`
- [x] `make generate` (tree clean)
- [x] `make test-unit` — new coverage in `pkg/middleware/openapi/remote/authorizer_internal_test.go` and `token_exchange_http_test.go`, updated assertion in `pkg/oauth2/passport_test.go`
- [ ] Re-run uni-region#531 against this branch to confirm the failing Integration spec passes

### DIFF
--- a/pkg/middleware/openapi/README.md
+++ b/pkg/middleware/openapi/README.md
@@ -81,14 +81,26 @@ passport claims payload, and the per-entry TTL is derived from the passport's `e
 10 s clock-skew fudge. Identity caps the passport expiry to the source token's expiry before
 minting it, so middleware does not need to parse the source token locally.
 
-The exchange path fails closed: rejected source tokens and transport failures surface as
-access-denied. A malformed or temporally invalid passport returned after a *successful* exchange
-response is treated as an internal failure (500) rather than access-denied, because that outcome
-reflects an identity-side defect, not a user authorization decision. Passport decoding rejects both
-expired (`exp` ≤ now) and not-yet-valid (`nbf` > now) tokens. There is no fallback to the legacy
-userinfo path. Passports are consumed in-process and are never forwarded on outbound calls —
-internal service-to-service communication continues to use mTLS plus `X-Principal` exactly as
-before.
+The exchange path fails closed and preserves the authentication vs. authorization distinction:
+
+- A 401 from the token endpoint indicates the subject token itself was rejected (missing, expired,
+  malformed, principal not active). The middleware surfaces this as `access-denied` (401) with the
+  generic "token is invalid or has expired" description, so authentication failures cannot be
+  distinguished from "unknown principal" — both look identical to a caller.
+- A 400 with `error: "invalid_scope"` indicates the subject token is valid but the principal is not
+  authorized for the requested organization/project scope. The middleware surfaces this as
+  `forbidden` (403) at the API edge so refresh-loop logic does not treat a wrong-scope request as a
+  bad-token request, and so client UX can distinguish "you need to reauthenticate" from "you can't
+  do that".
+- Transport failures and 5xx responses surface as upstream unavailability.
+- A malformed or temporally invalid passport returned after a *successful* exchange response is
+  treated as an internal failure (500) rather than an authorization decision, because that outcome
+  reflects an identity-side defect.
+
+Passport decoding rejects both expired (`exp` ≤ now) and not-yet-valid (`nbf` > now) tokens. There
+is no fallback to the legacy userinfo path. Passports are consumed in-process and are never
+forwarded on outbound calls — internal service-to-service communication continues to use mTLS plus
+`X-Principal` exactly as before.
 
 ## Ingress And Header Invariants
 
--- a/pkg/middleware/openapi/README.md
+++ b/pkg/middleware/openapi/README.md
@@ -81,21 +81,12 @@ passport claims payload, and the per-entry TTL is derived from the passport's `e
 10 s clock-skew fudge. Identity caps the passport expiry to the source token's expiry before
 minting it, so middleware does not need to parse the source token locally.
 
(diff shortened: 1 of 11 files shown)


pkg/oauth2/passport.go
```diff
@@ -310,7 +310,7 @@ func validateOrganizationScope(authz *openapi.AuthClaims, organizationID string)
 	}
 
 	if !slices.Contains(authz.OrgIds, organizationID) {
-		return errors.OAuth2AccessDenied("organization not in scope")
+		return errors.OAuth2InvalidScope("organization not in scope")
```
alansy-nscale, line 326: The comments above are now stale. It normalises the error to `invalid_scope` instead of `access_denied`.

pkg/oauth2/passport_test.go
alansy-nscale, line 1: Some tests in this file have not been updated to reflect the new status code mapping. Out-of-scope organization/project exchanges should now return 401 instead of 400.

pkg/middleware/openapi/README.md
```diff
@@ -81,14 +81,26 @@ passport claims payload, and the per-entry TTL is derived from the passport's `e
...
+- A 401 from the token endpoint indicates the subject token itself was rejected (missing, expired,
+  malformed, principal not active). The middleware surfaces this as `access-denied` (401) with the
+  generic "token is invalid or has expired" description, so authentication failures cannot be
+  distinguished from "unknown principal" — both look identical to a caller.
+- A 400 with `error: "invalid_scope"` indicates the subject token is valid but the principal is not
+  authorized for the requested organization/project scope. The middleware surfaces this as
+  `forbidden` (403) at the API edge so refresh-loop logic does not treat a wrong-scope request as a
+  bad-token request, and so client UX can distinguish "you need to reauthenticate" from "you can't
+  do that".
+- Transport failures and 5xx responses surface as upstream unavailability.
```
alansy-nscale: `ErrTokenExchangeUnavailable (500)` now falls through to `access_denied` in [this](https://github.com/nscaledev/uni-identity/blob/7be17adceaeb4cbb3ff133b1f0c3845c2775f600/pkg/middleware/openapi/remote/authorizer.go#L173) call path.

pkg/middleware/openapi/remote/token_exchange_http.go
```diff
@@ -85,23 +105,42 @@ func classifyTokenExchangeStatus(statusCode int) error {
...
+	return oauthErr.Error == oauth2ErrorInvalidScope
+}
+
 func decodeTokenExchangeResponse(resp *http.Response) (string, error) {
 	defer resp.Body.Close()
 
-	if err := classifyTokenExchangeStatus(resp.StatusCode); err != nil {
+	// Read once: the classifier needs the body on 400, the success path needs
+	// it for access_token, and the response stream can only be drained once.
+	body, err := io.ReadAll(io.LimitReader(resp.Body, errorBodySniffLimit))
```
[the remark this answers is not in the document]
alansy-nscale: @nscale-peter

alansy-nscale (changes requested):

pkg/middleware/openapi/remote/authorizer_internal_test.go
```diff
@@ -481,6 +482,70 @@ func TestCacheKeyedByTokenOnly(t *testing.T) {
...
+		{
+			name:          "forbidden sentinel from exchange becomes 403",
+			exchangeErr:   ErrTokenExchangeForbidden,
+			wantForbidden: true,
+		},
+		{
+			name:           "generic exchange failure falls back to access denied",
+			exchangeErr:    ErrTokenExchangeFailed,
+			wantAccessDeny: true,
+		},
```
alansy-nscale, line 519: The `ErrTokenExchangeUnavailable` case is missing from this mapping test.

alansy-nscale (approved):

alansy-nscale (approved):

PR #502, opened 2026-06-09
feat(oauth2): accept Auth0 access tokens at the local authorizer

alansy-nscale:
## Summary

Extends the local authorizer (used by every `/api/v1/...` endpoint) to accept Auth0 access tokens directly when Auth0 exchange is configured. The token-exchange endpoint already had this dispatch via `exchangeUserinfo`; this PR exposes that dispatch as a public method and threads it through the local authorizer, so callers presenting an Auth0 bearer no longer need a token-exchange round-trip first.

### Why

After #496 (v1.17.2) the token-exchange endpoint at `/oauth2/v2/token` can validate Auth0 access tokens. But the local authorizer at `/api/v1/...` still routes every bearer through `GetUserinfo`, which only decrypts JWEs. An Auth0 caller (the nscale console post-Auth0-migration) fails with `failed to parse jwe: compact JWE format must have five parts` because their RS256 source token is a 3-part JWS.

The exchange path's `isCompactJWS && auth0Validator != nil` dispatch is the right shape — this PR factors it into a public method and reuses it at the local authorizer so the same Auth0 bearer that works against `/oauth2/v2/token` also works against `/api/v1/...`.

### Changes

- `pkg/oauth2/passport.go`: new exported `GetUserinfoFromBearer` that dispatches compact-JWS tokens to the Auth0 validator and JWEs to `GetUserinfo`. Same decision logic as the private `exchangeUserinfo`, without the passport-source return value.
- `pkg/middleware/openapi/local/authorizer.go`: `authorizeOAuth2` now calls `GetUserinfoFromBearer` instead of `GetUserinfo` directly. One-line behaviour change with a comment explaining the dispatch.
- `pkg/oauth2/passport_test.go`: three regression tests covering the dispatch matrix.

### Behaviour notes

- **Fully backwards-compatible.** When `identity.auth0Exchange.issuer` / `.audience` are unset (chart default), `auth0Validator` is nil and the dispatch returns `GetUserinfo` verbatim. Deployments that haven't opted into Auth0 exchange see no change.
- For Auth0-sourced bearers, the resulting `claims.Type` is empty (the Auth0 path returns a `Claims` struct without a `Type`). The local authorizer's downstream switch in `authorizeOAuth2` therefore produces an `authorization.Info` with only `Token` and `Userinfo` populated — no `ClientID`, no service / service-account flags. The RBAC layer uses the userinfo's `Sub` + namespaced authz claim and does not depend on `ClientID` for the user path.
- UNI membership remains authoritative: `getAuth0Userinfo` ignores the Auth0 token's `orgIds` claim and queries `userdb.GetOrganizationIDs` by verified email (unchanged from #496).

## Test plan

- [x] `go test -count=1 ./pkg/oauth2/...` — three new tests plus the existing Auth0/UNI exchange suite all pass.
- [x] `go test -count=1 ./pkg/middleware/openapi/...` — clean.
- [x] `make touch license validate lint generate test-unit` locally.
- [ ] Smoke-test a deployment with `identity.auth0Exchange.issuer` / `.audience` set: present an Auth0 access token to `/api/v1/organizations` and observe a 200 instead of the prior 401 with `failed to parse jwe`.
- [ ] Confirm a deployment with `auth0Exchange.*` unset is unchanged: Auth0 bearers fail at the JWE decrypt step, UNI JWEs validate normally.


pkg/oauth2/passport.go
```diff
@@ -214,6 +214,21 @@ func (a *Authenticator) exchangeUserinfo(ctx context.Context, r *http.Request, t
...
 
+// GetUserinfoFromBearer dispatches to the Auth0 validator for compact-JWS
+// tokens and to the UNI userinfo path for JWEs. Used by the local
+// authorizer so Auth0 access tokens can be presented as bearers directly
+// to /api/v1/... without going through the token-exchange endpoint first.
+// When Auth0 exchange is not configured (auth0Validator is nil), every
+// token routes to the UNI userinfo path — preserving pre-Auth0-exchange
+// behaviour for deployments that haven't opted in.
+func (a *Authenticator) GetUserinfoFromBearer(ctx context.Context, r *http.Request, token string) (*openapi.Userinfo, *Claims, error) {
+	if isCompactJWS(token) && a.auth0Validator != nil {
```
spjmurray: Any attacker can craft a compact JWS and DoS auth0, probably blowing our rate limit and taking down the platform

alansy-nscale: Yeah, the concern is valid, but it already exists in #496 , which is now in dev. Could we get this through first so I can validate a few things, and then revisit the issue in a follow up?

spjmurray: Sadly that's not how security to the entire platform works, from a professional perspective.

alansy-nscale: After taking a second look, I realised we are not actually calling Auth0 [GET /userinfo](https://auth0.com/docs/api/authentication/user-profile/get-user-info). The `Userinfo` struct is built from claims already in the validated JWT + a UNI userdb lookup, so no outbound call to Auth0 happens once the token is verified.

That said, there is a real DoS vector on the JWKS path. When a `kid` is not cached, `go-oidc` refetches from Auth0, and only concurrent refreshes are deduped, so crafted tokens with random kids could each trigger a fetch and blow our rate limit. We could try mitigate this by adding a min refresh interval to the key set.

spjmurray: I'll buy into that!

spjmurray (commented):
Throttle looks good — nice call doing it at the HTTP transport layer rather than reimplementing the key set. The subtle bits are all handled: the window is consumed on failure too (so a failing/rate-limited Auth0 isn't hammered), the fetch timeout stops a hung fetch wedging go-oidc's `inflight` slot, and a fetch triggered under attack still refreshes the full JWKS so rotation isn't starved. Tests cover the right behaviours.

Two things I'd like to see:

**1. Observability on the throttle trip.**
Nothing emits a signal when a fetch is throttled, and we want it for two reasons: a spike in throttled fetches is exactly the attack signature we'd want to alert on, and we need to tell "throttle absorbing a spray" apart from "throttle rejecting a legitimate post-rotation token" — which currently surfaces as an indistinguishable `access denied`. Can we add a counter/metric on the throttled path, plus at most a *rate-limited* debug log (not per-request, or we'd self-inflict log spam under the exact attack this mitigates)?

**2. Routing hardening — and it applies to the exchange path too.**
Separate from the throttle but in the same area: the dispatch keys on token *shape* (`isCompactJWS` = `strings.Count(token, ".") == 2`). If Auth0 ever changes its access-token format — opaque, JWE, or a structural change, all within their rights — every Auth0 token silently misroutes to the UNI path and is rejected: a platform-wide auth outage for Auth0 users, no deploy, no warning. Suggested hardening:

- Route on the JOSE header instead of counting dots: a UNI token is a JWE (has an `enc` header) → UNI path; otherwise treat as a JWS → Auth0 validator (which already enforces iss + aud + signature, so a misroute can't cause wrongful *acceptance* — it's the silent *rejection* we're guarding against).
- Make "verifies as neither a valid UNI token nor a valid Auth0 token" a loud, metric'd rejection rather than a generic 401, so a format change shows up as an alert in minutes rather than a flood of mystery auth failures.
- A synthetic canary that logs in through Auth0 and asserts the token shape would catch a vendor format change before users do.

Note this dispatch is duplicated: `exchangeUserinfo` (the RFC 8693 token-exchange endpoint) and `GetUserinfoFromBearer` (this local-authorizer path) share the same `isCompactJWS` branch — the latter looks copied from the former. So the fragility, and any hardening, applies to **both**; worth factoring the dispatch into a single helper so the two can't drift.

Happy for the routing hardening to land as a follow-up if you'd rather keep this PR scoped to the throttle — flagging it here since this is what surfaced the dispatch.

alansy-nscale:
All of the suggestions above have been implemented.

spjmurray (approved):

PR #503, opened 2026-06-12
feat: Add GET /api/version endpoint

sravanmedarapu-work:
## Summary
- add authenticated `GET /api/version` returning the shared `coreapi.ServiceVersionRead` payload
- wire the OpenAPI spec/generated client/router and refresh the OpenAPI client mock for the new method
- update `github.com/unikorn-cloud/core` to `v1.17.1-0.20260612112435-4f6418b1367d`, the pseudo-version for core commit `4f6418b1367d` that defines `coreapi.ServiceVersionRead`
- add handler unit coverage and integration-suite coverage for authenticated and unauthenticated requests

## Verification
- `GOTOOLCHAIN=go1.25.8 make touch`
- `GOTOOLCHAIN=go1.25.8 make license`
- `PATH="$PATH:$(go env GOPATH)/bin" GOTOOLCHAIN=go1.25.8 make generate`
- `GOTOOLCHAIN=go1.25.8 make validate`
- `helm dependency build charts/identity`
- `PATH="$PATH:$(go env GOPATH)/bin" GOTOOLCHAIN=go1.25.8 make lint`
- `GOTOOLCHAIN=go1.25.8 make test-unit`
- `go test ./pkg/handler -run Test_Version_Get -count=1`
- `go test ./test/api -count=1`
- `go test -tags=integration ./test/api/suites -run "^$"`
- `go test ./pkg/client ./pkg/rbac ./pkg/handler -count=1`
- `golangci-lint run --timeout=10m --new-from-rev=origin/main ./...`
- `golangci-lint run --timeout=10m --build-tags=integration --new-from-rev=origin/main ./test/api/suites`
- `git status --short` clean after verification; only ignored coverage and Helm dependency artifacts were present

## Notes
- Base core PR: https://github.com/nscaledev/uni-core/pull/304
- This is not a `v1.17.1` release tag. No `uni-core` release is required for this change; the full pseudo-version points at the merged main commit and can be replaced with the next tagged core release when one is available.
- The unauthenticated integration case intentionally uses the low-level request helper so it can assert the raw 401 authorization failure path without trying to decode a version payload.


sravanmedarapu-work:
@claude please review the latest PR state, especially the core pseudo-version justification and verification notes.

JakeNScale (approved):
A clean, small change. Read the handler, the spec addition, both tests, and the auth middleware to confirm how `oauth2Authentication` is actually enforced. One real nit:

**`pkg/openapi/server.spec.yaml:57`** — nit: this endpoint declares a `403 forbiddenResponse`, but it can't produce one. The local authorizer only runs `authorizeOAuth2` for the `oauth2Authentication` scheme (→ `AccessDenied`/401 on bad or missing token), and since the path has no per-resource ACL, `GetACL` is never called — so 403 is dead in the wire contract. The closest precedent is `/oauth2/v2/userinfo` (also authenticated, also no RBAC), which declares only `401`. I'd drop `403` here to match it; the resource endpoints that legitimately declare `401 + 403` are the ones that actually do RBAC. (The `500` is fine — universally reachable.)

Everything else lines up: `setUncacheable`/`no-cache` and the zero-state handler match the package conventions, the file-level `//nolint:revive` mirrors the sibling handler files, and the integration test's `HavePrefix("unikorn-identity")` + `0.0.0`-or-`vX.Y.Z` assertions correctly track `constants.Application` (argv0 basename) and the Makefile's `VERSION=0.0.0` default.

Just a spec-accuracy nit — nothing blocking.

spjmurray:
Does it even need an /v2 prefix?  Version sounds as generic as can be, and can be used as a static discovery mechanism...

sravanmedarapu-work:
Updated this to use `GET /api/version` instead of `GET /api/v2/version`. Applied the same path shape across the sibling version endpoint PRs as well.

spjmurray (approved):

PR #507, opened 2026-06-17
feat(oauth2): resolve userinfo via bearer dispatch path

nscale-peter:
## Why

We want the v2 userinfo endpoint to handle different bearer token types. Previously it could only resolve UNI-issued tokens; now that the platform accepts Auth0 access tokens at the local authorizer, the same endpoint needs to transparently resolve both UNI and Auth0 bearers without callers having to know which they hold.

## What

Routes the v2 userinfo handlers (`GetOauth2V2Userinfo` and `PostOauth2V2Userinfo`) through `GetUserinfoFromBearer` instead of `GetUserinfo`, so the local authorizer resolves both Auth0 and UNI bearer tokens via the shared `dispatchUserinfo` path. UNI access tokens are JWEs and continue to flow through the existing `GetUserinfo` verification unchanged, so the happy path for UNI-token users is byte-identical. The only behavioural delta is at the edges: malformed/non-JOSE tokens are now rejected earlier as "unrecognized bearer token format" and counted on the `unroutableTokens` metric, turning a token-format regression into an alertable signal. This also unifies the bearer and token-exchange entry points so the two cannot drift.


spjmurray (approved):
Begrudging approval, but I do have your word that you will do local auth0 validation and just read the email, out from the token as soon as practicable and where necessary.

PR #508, opened 2026-06-17
feat: scope-reader interfaces for cleaner, type-safe scope handling

spjmurray:
## Why

Services that own project-scoped CRDs (region, …) repeatedly do the same dance before calling identity's RBAC, principal-enrichment, and reference APIs: read the org/project off the resource's labels, parse them to typed IDs, then pass the IDs across. That boilerplate is duplicated at every callsite and is easy to get subtly wrong. The same pattern shows up *inside* identity's own handlers, where placement labels are set with stringified IDs.

This change defines the interfaces those resources already satisfy and lets the consumer **pass the resource itself** (or its typed IDs), so the "read labels → parse to typed ID → stringify" steps happen **once, in the right layer**, instead of at every call. The headline goal is cleaner consumer code with type safety preserved as far up the stack as possible.

## What

- **`pkg/ids`** — new `OrganizationScopeReader` and `ProjectScopeReader` interfaces (the latter embeds the former: a project always belongs to an organization), plus tenancy comparison helpers over them: `SameProject(a, b)` (do two resources share an org+project?) and `OwnedByProject` / `OwnedByOrganization` (does a resource live in this org/project?). The latter pair are typed equivalents of core's `Assert*Ownership` — referential-integrity equality checks, deliberately **not** RBAC (RBAC would let a multi-project caller relate resources across tenancy boundaries).
- **`pkg/rbac`** — additive `…Reader` variants (`AllowOrganizationScopeReader`, `AllowProjectScopeReader`, `AllowProjectScopeCreateReader`) that recover IDs from the resource and delegate to the existing typed `…ID` variants.
- **`pkg/principal`** — `EnrichUserPrincipal{Organization,Project}Scope{ID,Reader}`, the typed home for principal enrichment that previously lived as a bespoke helper downstream. Like the RBAC helpers, the `…ID` variants take typed IDs (path params) and the `…Reader` variants take a CRD.
- **`pkg/client/references.go`** — recovers IDs via `ProjectScopeReader` when the resource implements it, falling back to label parsing.
- **`pkg/handler/common`** — `SetIdentityMetadataOrganizationScope` / `SetIdentityMetadataProjectScope` stamp the placement organization/project labels from typed IDs and set the attribution metadata in one call. Identity's create paths (project, group, quota, service account, oauth2 provider, organization user, allocation) are migrated onto them.

## Type safety, preserved as long as possible

The scope checks come in three argument flavours so callers pass whatever they already hold and IDs stay typed end-to-end, with string conversion deferred to the lowest layer:

- `…` (string) — **deprecated**, retained for backwards compatibility while callers migrate.
- `…ID` (typed `ids.*`) — for handlers holding decoded path parameters.
- `…Reader` (resource) — for callers holding a CRD.

The string variants are marked `// Deprecated:`; staticcheck SA1019 only flags cross-module use, so this nudges external consumers to migrate without nagging identity's own code.

## Moving client hacks back where they belonged

- `pkg/client/references.go` carried a local label-read-and-parse helper — exactly the logic that should live behind a shared contract. It now goes through `ProjectScopeReader`, with the resource's own accessor doing the work, and `SameProject` lives in `pkg/ids` rather than being re-implemented per consumer.
- Identity's handlers set placement labels with `conversion.NewObjectMetadata(...).WithOrganization(id.String())`. Core's builder is necessarily string-based — core cannot import `identity/pkg/ids` without a circular dependency — so the typed equivalent can only live in identity. `SetIdentityMetadata{Organization,Project}Scope` provide it, keeping the typed ID to the lowest layer and stamping placement + attribution together so they can't drift apart.

## Backwards compatibility

Everything here is backwards compatible:

- The string and `…ID` variants are unchanged; the `…Reader` and `SetIdentityMetadata*Scope` variants are purely additive (bare `SetIdentityMetadata` stays for org-less resources).
- `references.go` falls back to label parsing for resources that have not yet adopted the accessors — both paths read the same labels and agree by construction.
- Migrated handlers write identical label keys and values; no API/OpenAPI surface changes.

## Follow-up (region, separate PR)

- Add `OrganizationID()` to region's project-scoped CRDs so they satisfy the embedded `ProjectScopeReader`.
- Migrate region's local `InjectUserPrincipal` to `principal.EnrichUserPrincipalProjectScopeReader` and adopt the `…Reader` RBAC variants where a CRD is in hand.

### DIFF
--- a/pkg/client/README.md
+++ b/pkg/client/README.md
@@ -81,6 +81,11 @@ This is an important forward-looking deletion mechanism because it is based on
 explicit dependency registration rather than only namespace-local descendant
 discovery.
 
+`References` recovers the typed organization and project IDs that own the target
+resource via [`ids.ProjectScopeReader`](../ids/README.md) when the resource
+implements it (region CRDs do), falling back to parsing the standard
+organization/project labels otherwise.
+
 ### Allocations
 
 `Allocations` wraps the identity quota/allocation API used by API handlers and
--- a/pkg/client/export_test.go
+++ b/pkg/client/export_test.go
@@ -32,6 +32,11 @@ func ParsePrincipalIDs(p *principal.Principal) (ids.OrganizationID, ids.ProjectI
 	return parsePrincipalIDs(p)
 }
 
+// OrganizationAndProjectID exposes organizationAndProjectID for unit testing.
+func OrganizationAndProjectID(resource crClient.Object) (ids.OrganizationID, ids.ProjectID, error) {
+	return organizationAndProjectID(resource)
+}
+
 // SetClientFactory replaces the HTTP client factory on r, allowing tests to
 // inject a mock without going through the full Kubernetes service-discovery path.
 func (r *References) SetClientFactory(f func(ctx context.Context, c crClient.Client, resource crClient.Object) (openapi.ClientWithResponsesInterface, error)) {
--- a/pkg/client/references.go
+++ b/pkg/client/references.go
@@ -36,9 +36,13 @@ import (
 )
 
 // organizationAndProjectID recovers the typed organization and project IDs that own a
-// resource. If the resource implements ids.ProjectScopeReader (region CRDs do) the typed
-// accessor is used directly; otherwise it falls back to parsing the standard organization
-// and project labels.
+// resource. If the resource implements ids.ProjectScopeReader the typed accessor is used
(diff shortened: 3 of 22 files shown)


spjmurray:
@claude review

alex-emery (approved):

PR #512, opened 2026-06-30
feat: multi-issuer bearer auth with eligibility/authority split (ID-284)

andreyyantsen-nscale:
## What & why

uni-identity's token validator was pinned to a single OIDC issuer. This change lets it trust a
**configurable list of issuers** — each pinned to its own JWKS and verified locally — dispatch an
incoming bearer token to the matching issuer by an **exact `iss` match**, and resolve every token
through the **same RBAC entry point**, with authority conferred only by a server-side grant, never
by a token claim.

Motivating use case: let internal/staff identities (a second IdP tenant) reach the same API as
customer identities, without a bifurcated auth path and without modelling staff as an organization.
This is a **read-only, mechanism-only** change.

## What's in it

- **Multi-issuer JWT verification.** A `bearerTrust` block on the `OAuth2Provider` CRD opts an issuer
  into bearer trust (audience, signing-algorithm allowlist, email/authz-claim policy). A per-provider
  validator (cached by name + spec fingerprint) verifies signature, issuer, audience and temporal
  claims. The issuer is matched **verbatim** (exact string, per OIDC Core §3.1.3.7) — no
  normalization — so a token from any configured issuer validates and all others are rejected.
- **One RBAC path for every issuer.** Tokens from either tenant resolve through the same
  `processUserAccountACL`; there is no second code path. The verbatim source issuer (`src_iss`) is
  carried through `authorization.Info` and stamped on the minted passport.
- **Eligibility vs authority.** A valid token from a trusted issuer *resolves to a principal*
  (eligible), but authority requires a server-side grant: the platform-admin match is on the
  `(src_iss, subject)` pair, closing the cross-issuer confused-deputy and staying independently
  revocable. A group/role claim never confers authority on its own.
- **Global read without org membership.** A principal carrying a global ACL returns before the
  org-membership check, so `GET /organizations` lists all orgs for a platform admin.
- **No staff "org".** External identities with no local membership carry an empty (non-null)
  `orgIds`, so staff are not modelled as an organization and don't leak into cross-org queries.

A startup migration gate refuses to boot a bare (single-issuer) admin list once a second issuer is
trusted.

## Out of scope

Signed-attestation propagation, long-lived signing authority + key rotation, failure/rebind,
break-glass, customer-as-approver, and impersonation / write powers. Read-only only.

## Testing

`make test-unit` (unit coverage across `pkg/oauth2` and `pkg/rbac`, including verbatim-issuer
matching and the eligibility/authority split) plus integration coverage in
`test/api/suites/bearer_trust_test.go`. Migration-gate startup behaviour is covered in `pkg/server`
and `pkg/rbac`.


spjmurray (approved):
After a very long chat with my AI friend, we agree that this brings real value, but in the wrong place for technical reasons I've discussed with you.  I approve in the proviso you treat those technical reasons with due attention!

spjmurray (approved):

PR #518, opened 2026-07-21
feat(authz): generate Cerbos policies from Role CRDs

alansy-nscale:
## Summary

First PR in a **proof of concept** for moving identity's authorization *decisions* onto a Cerbos policy engine, motivated by several goals of roughly equal weight: **attribute-based access control (ABAC)** — decisions that depend on resource attributes and request context (classification, ownership, step-up/MFA, source IP) that today's role-only engine can't express — **policy-as-config**, **decoupled authN/authZ**, and **reusable, auditable decisions**. The PoC reproduces today's RBAC behaviour first (shadow-validated, cut over per kind), so Cerbos can be evaluated and adopted incrementally and reversibly. Full design: `docs/cerbos-authorization-design.md`.

This PR adds **only the policy generator**: a pure, offline transform from identity `Role` CRDs to a Cerbos policy store. There is no runtime wiring, no sidecar, and no decision path yet — those arrive in later PRs in the stack, so this one is safe to land on its own.

## What's in this PR

- **`pkg/authz/cerbos/generate`** — compiles the `Role` CRDs into a Cerbos store: **derived roles** plus **per-resource policies**, carrying the global → organization → project scope hierarchy.
- **Golden-file tests** (`generate_test.go` + `testdata/store/**` and `testdata/store/tests/**`) that pin the generated output and the Cerbos policy test suite, so any change to generation is visible in review.
- **`docs/cerbos-authorization-design.md`** — the design note that sets context for the whole series (motivations, scope hierarchy, rollout approach).

## Deliberately not here (later in the stack)

- Cerbos PDP sidecar + fail-closed gRPC client
- Reconciliation controller that publishes the store into a ConfigMap at runtime
- Decision API, dual-path engine seam, shadow mode, `/authorization/check` endpoint, downstream remote path, front-door audit
- The region / compute / kubernetes service-account supersets (the generated store here is the baseline; those roles' scopes are added in the final PR)
- ABAC condition predicates — this lands the role-equivalent policies first; the attribute- and context-based predicates (one of the PoC's goals) come later

## Testing

- `go build ./...`
- `go test ./pkg/authz/cerbos/generate/...` — golden-file tests pass
- The `cerbos compile` gate over the generated store (`make validate-policies`) lands with the pinned sidecar image in a later PR

## Notes

- **Base of a stacked, PoC-first series**, targets `main`; subsequent PRs stack on top of this branch.
- The generator is deterministic and side-effect-free — it reads `Role` CRDs and emits YAML; nothing in this PR runs at request time.


spjmurray:
Nice first step — the offline generator plus golden-file pinning is a clean, safe way to land this, and reproducing today's RBAC first (shadow-validated, per-kind) is the right de-risking. No concerns with this PR standing on its own.

One directional consideration worth raising early, because it shapes how the rest of the stack is sequenced rather than anything in this PR:

As the platform moves toward regionally-distributed control planes, authorization will increasingly need to be **evaluated locally within each region** rather than via a per-request call to a central PDP. Part of that is latency and availability — keeping the decision off a cross-region round-trip, and letting a region keep serving while temporarily isolated from central services — but there's also a security case specific to a policy engine:

- **Attribute egress.** ABAC decisions consume resource attributes and request context (classification, owner/team, MFA state, source IP). A central PDP ships all of that out of the region on every check; local evaluation keeps those attributes where the resource lives — and keeps the policy engine from becoming a copy of every service's data.
- **No fail-open temptation.** A central PDP forces a binary on every blip: fail-closed (deny → unavailable) or fail-open (allow → insecure). Local evaluation against a bounded last-known-good policy replaces that with a principled freshness window — a region stays available without loosening deny-by-default.
- **Contained blast radius.** Concentrating every decision in one network-reachable engine concentrates the trust; local per-region evaluation keeps a bad bundle or a compromised engine scoped to one region rather than every service at once.

That bears on two choices the design currently makes:

1. **Central (A) vs per-service/regional (B).** The design treats distributed PDPs as "deferred until capacity." Given the above, it's worth treating local evaluation as the *target* and shaping the M1 seams for it from the start (they largely are) — so B is a rollout, not a re-architecture.

2. **The §10.1 "single enforcement point / no local policy evaluation" assumption.** The design reasonably reads this as requiring central today and notes B would need an amendment. That amendment looks foundational to the distributed direction rather than incidental, so it's probably worth opening the spec conversation now. A useful reframing: "single enforcement point" as a single policy *authority* — one place that owns and signs policy — *evaluated locally everywhere* against a distributed copy. One source of truth, not one evaluation location.

The good news: the artifact in this PR — a generator plus a reconciling controller that publishes a policy store the PDP hot-reloads — is topology-agnostic, and is exactly what you'd want for distributing policy out to regional PDPs. None of this blocks the PR; it's about keeping the later stack pointed at the distributed end-state.

Two things to keep in view when the engine choice firms up:

- Weigh the policy-distribution-to-many-PDPs story (signed bundle distribution / hot-reload to N regional PDPs), not only the central footprint — the candidates differ there.
- Signed bundles pushed to regional PDPs give each region cryptographic assurance of exactly which policy it's enforcing, plus a verifiable policy-version-per-decision in the audit log — which complements the controller-integrity work already here (compile-gate, least-privilege ConfigMap, content-hash keys).

PR #528, opened 2026-07-31
fix(handler/groups): validate role list changes as a delta (ID-368)

nscale-peter:
Second of four PRs splitting the group-permission work ([ID-368](https://linear.app/nscale-workspace/issue/ID-368/clarify-permission-management-for-group-membership)). **Stacked on #527** — review that first; this PR's base retargets to `main` automatically when it merges.

This PR fixes the reported symptom: org admins could not edit a group that carried a role they were not allowed to grant.

## Problem

Group create and update grant-checked the **entire** role list in the request body, not just what the request changed. A group carrying a role from another service (radar, reservations) therefore could not be renamed, and its members could not be edited, by anyone who could not grant that role — even though the request left the role list untouched. That is what the NVIDIA reports were hitting.

The mirror problem is that dropping such a role was completely unguarded. A client that fetched the roles it could see and sent that list back removed the role it could not see, silently.

## Change

- **Additions are checked; re-sends are not.** `validateRoleIDs` now takes the group's stored role list and grant-checks only roles the request adds. Create still checks everything, since every role on a new group is an addition. Existence and `protected` checks continue to cover the whole list.
- **Removals are guarded.** Dropping a role the caller cannot grant is refused with a 403 naming the role. Two exemptions, both erring towards *less* authority: a role reference that no longer resolves (nothing to revoke, and dropping it is how a group carrying one gets repaired), and a `protected` role (it should never be on a group, so removing one is invariant repair).
- **Refusals name the role** — display name and ID — via a new shared `common.RoleDisplayName` helper, so callers can act on the error.
- **CI covers roles outside the built-in set.** The grant-lattice guard test now parses the chart's `additionalRoles`, so an operator-installed role that no built-in role can grant fails the build rather than surfacing as a support ticket.
- **Integration harness.** A Kubernetes client for fixtures that have no HTTP API (Role and Group CRs), typed group-update wrappers on the test client, and the `IDENTITY_NAMESPACE` plumbing the suites need.

Group membership is deliberately untouched here — adding a member to a group is still ungated, exactly as on `main`. PRs 3 and 4 in this series change that.

## Compatibility

No CRD changes, no schema changes, no migration. One behaviour change worth a release note: **an update that omits a role the caller cannot grant now fails with 403 instead of silently revoking it.** Automation that round-trips a filtered role view will start failing loudly. That is the intent — but it will surface as new errors on existing deployments, and the fix for such a client is to send the role back unchanged.

## Testing

- Unit: the delta matrix (create checks all, update checks additions, re-sends pass, genuine additions still refused, protected always refused); the removal guard directly and through `Update`, including the dangling and protected exemptions.
- Integration, against a live deployment: with an ungrantable role fixture installed by direct CR access, an unchanged re-send of the group's role list succeeds and the role survives; dropping it is refused with the role named. Assertions match the guard-specific error text rather than just the role name, since several guards name roles.
- Full suite measured against a `main` baseline on the same cluster: same failure set, none attributable to this change. Those pre-existing failures are an environment problem with the `USER_AUTH_TOKEN` fixture, not the code.


spjmurray:
Reviewed alongside #529 and #530, since the three only make sense together.

Fine to land, on the condition it does not ship without #530. Confirming why, because the description understates it.

## The description is wrong about membership

> Group membership is deliberately untouched here — adding a member to a group is still ungated, exactly as on `main`.

Not for this endpoint. On `main` a group carrying an ungrantable role could not have its member list edited at all — the whole-list check refused every write to it. Checked with the package's own fake-client fixtures: caller holds only `identity:groups update`, group carries a role scoped to `radar:things`, request re-sends the role and adds a member.

- `main`: refused, `requested role cannot be granted`, group unchanged.
- this branch: accepted, the new member now holds `radar:things`.

It does not stop at that group either. Once the ACL refreshes the new holder passes `AllowRole`, so they can attach the role to fresh groups via `Create`. Also verified.

None of that is new to the system — `PUT /users` with `spec.groupIDs` has always done the same with no check at all, which is what #530 fixes. But it is new to this endpoint, and "exactly as on `main`" is the sentence that hides it. Please reword, so the next reader can see why #530 has to land with it.

## Drop the removal guard

`validateRoleRemovals` refuses dropping a role you cannot grant. Dropping a role cannot escalate anybody, and the guard is walkable three ways: delete the group, empty the member list, or omit a dangling role.

It also creates a new stuck state, the one the README spends thirty lines documenting: once nobody holds a role's permissions, nobody can remove it through the API. That is the support ticket this PR set out to kill, moved rather than fixed.

Keep the additions delta — that is the right model and it fixes a real bug. Bin the removal guard and the compatibility release note goes with it.

## TestNonBuiltinRolesAdminGrantable

Two problems. It asserts every non-protected role must be grantable by `administrator`; if that held, the bug this PR fixes could not exist. And it reads this chart's `values.yaml`, so it can never see a role shipped by radar or any other service, which is where these roles actually come from. Delete it, or make it a runtime check when a `Role` CR appears.

## The integration test probes with the thing under test

`waitForFixtureVisibility` uses a real `UpdateGroup` re-send as its cache probe, and that succeeding is exactly what the first `It` claims to verify. Revert the production change and the suite fails in `BeforeEach` saying the role fixture is not visible yet, which blames the cluster. Poll the `Role` CR through the kube client the spec already holds instead.

## Minor

- `validateRoleIDs` and `validateRoleRemovals` each `Get` the same `Role` objects and repeat the not-found / protected / `AllowRole` chain. One pass over added/kept/removed is smaller and puts the whole rule in one place.
- `test/api/suites/group_membership_test.go` tests role deltas, not membership, and will collide with #529's actual membership specs.

## The underlying problem

Worth naming this as a workaround. `administrator` cannot grant `radar-admin` because it does not hold `radar:*`. That is a grant-lattice problem: a service shipping a tenant role has no way to declare who may grant it. The delta makes the symptom go away without touching that.

Related, and feeding the same confusion: `protected: true` is doing two jobs. It covers `region-service` and friends (machine identities that should never sit on a group) and `platform-administrator` / `platform-reader` (staff roles that are fine on a group, just not for tenants to hand out). One flag, two rules, so the strict one wins everywhere — which is why a platform admin cannot attach `platform-reader` through the API, and why `globalGroupRoleBindings` grew a path that skips the flag. A `kind` enum on the `Role` type would cover it without new CRDs and the boilerplate that entails.

spjmurray (approved):
All four points addressed, and the rewritten description is a better account of the change than the original. Approving.

- removal guard gone entirely, and `validateRoleIDs` now says why removals are ungated
- `TestNonBuiltinRolesAdminGrantable` dropped, `grant_guard_test.go` back to `main`
- visibility probe is reads only, and the file rename to `group_role_delta_test.go` frees the membership name for #529
- the "This opens a membership window that #529 and #530 close" section is the important addition — it means the next reader cannot merge this without knowing what it depends on

Two things before you merge.

**Fold the three `(review)` commits.** We rebase rather than squash, so they land on `main` verbatim. Fold each into the commit it fixes, so the history reads as if it were right first time.

**Do not cut a release between this and #530.** Merging the stack in order is fine; tagging in the middle of it is not. #529 closes the groups-path window this opens, #530 closes the `PUT /users` and service-account paths that were never gated at all.

PR #529, opened 2026-07-31
feat(handler/groups): gate member additions on the group's roles (ID-368)

nscale-peter:
Third of four PRs splitting the group-permission work ([ID-368](https://linear.app/nscale-workspace/issue/ID-368/clarify-permission-management-for-group-membership)). **Stacked on #528** — review #527 and #528 first; the base retargets as each merges.

## The rule this implements

A group carries roles, and every member of the group inherits them. Adding a member therefore *confers* those roles, so it is a grant like any other and must trace to a holder: the caller has to be able to grant every role the group carries. Removals, principal deletion and group DELETE take authority away rather than handing it out, and stay ungated.

This is the maintainer ruling from review — no actor confers authority on itself, and every grant traces to a prior holder.

## Change

- **`PUT /groups/{id}` gates member additions** on the group's role list. A refusal names the blocking role.
- **Members are compared by principal identity, not by struct equality.** A subject is `(Issuer, ID)`; `Email` is display data that the writers of these records populate from different sources, so it takes no part in identity. Presence in *either* the legacy `UserIDs` list or `Subjects` counts as already-a-member, because RBAC resolves membership through both — completing the other half confers nothing and is not a grant. Without this, a group still using the legacy representation reads every write as adding its existing members, and a no-op re-send or a member *removal* would be refused by the addition gate.
- **Two exemptions that confer nothing:** a group with no roles, and a role reference that no longer resolves — the latter refuses the addition rather than skipping it, because role IDs are derived from the role name and a role that is deleted and re-applied re-binds to every group still referencing it.
- **Create needs no membership check.** Every role on a new group is an addition and is already grant-checked, so the creator holds everything the group confers.

Scope: this PR gates the groups path only. `PUT /users/{id}` and service-account create/update remain ungated here, exactly as on `main`; PR 4 closes those and is where the "same rule everywhere" claim is earned.

## Compatibility

No CRD schema changes and no migration — `group_helpers.go` adds methods to existing types, so the generated CRDs and deepcopy are untouched.

One behaviour change worth a release note: **on deploy, a group carrying a role its editors cannot grant becomes frozen for member additions.** Removals, renames and role-list re-sends continue to work. The 403 names the blocking role. The unblock for an affected team is a holder of that role granting it to whoever administers the group.

There is also an interpretation change on unchanged data: a principal recorded only in the deprecated `UserIDs` list now counts as already a member of that group. Nothing stored is rewritten.

## Testing

- Unit: the gate proved non-vacuous in both directions — forcing the guard off fails the four rejection tests, forcing it on fails the four allow tests. Cases cover legacy `UserIDs`-only groups (no-op re-send allowed, genuine addition refused), a stored subject whose email differs from the derived one, subjects-only storage, an external subject riding in alongside an existing member, a roleless group, and an addition by a caller who does hold the role.
- Integration, against a live deployment: addition refused naming the role; removal allowed with the role surviving; and a positive control where the role IS grantable and the addition succeeds — without that control, a guard that refused everything would still look green.
- Full suite measured against a `main` baseline on the same cluster. The delta is entirely the `USER_AUTH_TOKEN` fixture, which is broken on that cluster independently of any branch; all five new specs pass inside the full run.

### DIFF
--- a/pkg/apis/unikorn/v1alpha1/group_helpers.go
+++ b/pkg/apis/unikorn/v1alpha1/group_helpers.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2026 Nscale.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"slices"
+)
+
+// IdentityKey returns a comparable key for the principal a subject names.  A
+// principal is identified by the issuer that authenticated it and its ID at
+// that issuer.  Email is display data only — the handlers that write subjects
+// populate it from different sources and it can differ between two records
+// for the same principal — so it takes no part in identity.
+func (s *GroupSubject) IdentityKey() string {
+	return s.Issuer + "\x00" + s.ID
+}
+
+// Matches reports whether two subjects name the same principal.
+func (s *GroupSubject) Matches(other GroupSubject) bool {
+	return s.IdentityKey() == other.IdentityKey()
+}
+
+// HasSubject reports whether the group already lists a subject for the same
+// principal.
+func (s *GroupSpec) HasSubject(subject GroupSubject) bool {
+	return slices.ContainsFunc(s.Subjects, subject.Matches)
+}
+
+// HasMember reports whether the group already confers its roles on the
+// principal.  Either membership representation counts: RBAC resolves a
+// principal's groups through the subject list and through the deprecated
+// organization user ID list alike, so presence in one is enough for the
+// principal to hold the roles already, and writing the other half confers
+// nothing.
+//
+// An empty organization user ID matches nothing.  Membership lists are not
+// validated against real records, so a junk empty entry must not stand in for
+// a principal that has no organization user record yet.
+func (s *GroupSpec) HasMember(organizationUserID string, subject GroupSubject) bool {
+	if organizationUserID != "" && slices.Contains(s.UserIDs, organizationUserID) {
+		return true
+	}
+
+	return s.HasSubject(subject)
+}
--- a/pkg/handler/common/membership.go
+++ b/pkg/handler/common/membership.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2026 Nscale.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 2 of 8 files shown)


spjmurray:
The gate is the right design and it does what it claims. Checked with the package's own fake-client fixtures: the member addition that #528 opened up is now refused.

```
members cannot be added to the group: role "radar" (radar-id) is not grantable by the caller
```

Everything I tried to get round it failed closed:

- dropping the ungrantable role and adding a member in one write is allowed, but the role is gone, so nothing is conferred
- dropping it, then re-adding it, is refused by `validateRoleIDs` — the re-add is an addition
- `Create` needs no gate, since every role on a new group is grant-checked already
- a dangling role reference refuses the addition, and the repair path still exists because removals are ungated

`HasMember` is stricter than RBAC in both its branches, so it never fails open. It does fail closed too often, though.

## The gate uses the wrong membership predicate

`hasMemberAdditions` asks `current.Spec.HasMember(principal.userID, principal.subject)`, which matches subjects on issuer **and** ID. `groupSubjectFilter` in `pkg/rbac` matches on ID alone, deliberately, because subject records written before issuers were recorded carry an empty one. So a legacy record already confers the group's roles, but the gate reads a write that re-states it as an addition and refuses it.

Group with `subjects: [{id: alice@example.com, issuer: ""}]` and no `userIDs`, admin who cannot grant the group's role, request re-sends the membership as `userIDs`:

```
members cannot be added to the group: role "radar" (radar-id) is not grantable by the caller
```

Nothing is being conferred there — RBAC already resolves alice into that group. That is the ID-368 symptom coming back through the new door.

You already found this on the users path in #530 and added `HasMemberByID` for it, whose doc says outright that a gate asking whether the group's roles are already conferred must use it. The groups call site did not move with it. Ported that one line back onto this branch:

```go
if !current.Spec.HasMemberByID(principal.userID, principal.subject.ID) {
```

The over-refusal goes away, the escalation stays refused, `go test ./pkg/...` still green.

Worth noting what that does to the pair: in #530, `HasMember` has exactly one caller, this one. Fix it and the method is dead — `Matches` and `HasSubject` are still used by `addToGroup`, but `HasMember` is not. So this is one predicate, not two, and the issuer-qualified variant can go.

## Unrelated

#530 is based on this branch, which was force-pushed onto the reworked #528, so #530 currently carries `validateRoleRemovals` and the deleted grant-guard test back in. It needs rebasing before it is reviewable.

nscale-peter:
Fixed in `a7d2dbaf`. The gate now uses `HasMemberByID` — subject matched by ID alone, mirroring `groupSubjectFilter` — so a legacy empty-issuer member re-sent as a `userID` no longer reads as an addition. The issuer-qualified `HasMember` is gone, along with the now-unused `HasSubject`/`Matches`; `IdentityKey` stays for stored-subject dedup (issuer-qualified there on purpose). Added a unit test for that exact case: it fails against the old predicate with the over-refusal you hit and passes now. README and PR description updated to say ID-only.

#529 has been rebased onto merged #528 (`e762c770`) and force-pushed, so it's clean against `main`. #530 next: I'll rebase it onto this and drop the `validateRoleRemovals` + deleted grant-guard test drift (and the now-duplicate `HasMemberByID`).

spjmurray (approved):
Fix is right, and deleting `HasMember`/`HasSubject`/`Matches` rather than leaving two predicates side by side is the better answer — one question, one method. Keeping `IdentityKey` for stored-subject dedup is the correct split, and the comment now says why the two differ.

Re-ran the checks on `a7d2dbaf`:

- the member addition #528 opened is still refused — `members cannot be added to the group: role "radar" (radar-id) is not grantable by the caller`
- the legacy empty-issuer member re-sent as a `userID` is now accepted
- a genuinely new external subject alongside an existing member is still refused, so matching by ID alone did not open anything
- `go test ./pkg/...` green

Everything I tried to walk round the gate failed closed: dropping the ungrantable role and adding a member in the same write leaves nothing conferred; dropping then re-adding it is refused by `validateRoleIDs`; `Create` needs no gate because every role on a new group is checked; a dangling role reference refuses the addition and removals stay ungated so the repair path survives.

Approving. Same condition as #528 — no release until #530 lands, since `PUT /users` and the service-account path are still ungated until then.

PR #530, opened 2026-07-31
feat(handler): gate membership additions on the user and service-account paths (ID-368, ID-367)

nscale-peter:
Fourth and final PR splitting the group-permission work ([ID-368](https://linear.app/nscale-workspace/issue/ID-368/clarify-permission-management-for-group-membership)). **Stacked on #529** — review #527, #528 and #529 first; the base retargets as each merges.

This one also closes acceptance criterion 1 of [ID-367](https://linear.app/nscale-workspace/issue/ID-367/nvidia-bug-revoke-elevated-access-after-role-downgrade).

## Problem

#529 established that adding a member to a group confers that group's roles, and gated it — on the groups path. Two other paths write group membership and had **no grant check at all**:

- `PUT /users/{id}` reconciles a requested `groupIDs` list into membership.
- Service-account create and update do the same, and a service account carries a long-lived bearer token whose authority is independent of whoever created it. A user whose own authority is being reduced could mint one attached to a group they could not grant, and the token outlives the downgrade.

Both also wrote before they authorized: a refused request had already persisted its other changes.

## Change

- **The same gate on both paths.** Every group the principal would newly join is checked against the caller's authority; a refusal names the blocking role. Removals are not gated, and principal deletion is exempt — it strips memberships as cleanup, and gating it would make a principal sitting in a group with ungrantable roles undeletable.
- **Validate then apply.** Authorization for a multi-group write is settled before the first patch, so a refusal applies nothing.
- **Validate before persisting.** `users.Create` no longer writes the global user or organization user, and service-account create no longer creates the account or issues its token, until authorization has passed. Both paths are idempotent and reuse existing records, so the pre-flight resolves what the principal already has without creating it and exempts memberships they already hold — a legitimate re-create is not refused.
- **Unknown group IDs are refused.** A requested group that does not exist in the organization was silently dropped from the write and reported as success; it is now an explicit error.
- **Service-account responses report the groups actually joined**, rather than a pre-write snapshot.
- **`pkg/rbac/handler.go`'s trust note is corrected.** It justifies the grant lattice's laxest rule — accepting a project-scoped permission held in *any* of the caller's projects — on an enumeration of every route to `AllowRole`. This stack adds routes; the enumeration is now complete and each entry's organization-scope check is named, so the recorded security reasoning is true when the stack lands.

## Compatibility

No CRD schema changes, no migration.

Release notes:

- **Member additions on `PUT /users/{id}` and service-account create/update now require the caller to hold the group's roles.** These paths were unguarded, so any org relying on that will see new 403s. The error names the blocking role.
- **An unknown group ID in a request is now 400 rather than a silent success.** This is a compatibility break beyond the security fix and is called out deliberately: clients previously got a 200 and a response body that quietly omitted the group.

## Testing

- Unit: 23 new subtests across users, service accounts and the shared helper — refusal named, removal allowed, roleless group allowed, holder allowed, multi-group atomicity (nothing applied on refusal), no principal or service account persisted when a create is refused, idempotent re-create still permitted, unknown group refused.
- Integration, against a live deployment: joining a group with an ungrantable role through the users API is refused naming the role; leaving it is allowed.
- Full suite measured against a `main` baseline on the same cluster: the branch's failures are a strict subset of the baseline's — zero new failures. Those baseline failures are a broken `USER_AUTH_TOKEN` fixture on that cluster, unrelated to any branch and worth its own ticket.

## Known limitation, stated rather than hidden

These paths are atomic against *refusals* — authorization is fully settled before the first write — but not against infrastructure failures partway through a multi-write request. That is pre-existing and now documented in the package READMEs rather than left implicit.

### DIFF
--- a/pkg/README.md
+++ b/pkg/README.md
@@ -104,6 +104,7 @@ This is most visible in:
 
 - [handler](./handler/README.md)
 - [handler/users](./handler/users/README.md)
+- [handler/serviceaccounts](./handler/serviceaccounts/README.md)
 - [handler/groups](./handler/groups/README.md)
 - [handler/quotas](./handler/quotas/README.md)
 - [handler/allocations](./handler/allocations/README.md)
--- a/pkg/apis/unikorn/v1alpha1/group_helpers.go
+++ b/pkg/apis/unikorn/v1alpha1/group_helpers.go
@@ -50,10 +50,42 @@ func (s *GroupSpec) HasSubject(subject GroupSubject) bool {
 // An empty organization user ID matches nothing.  Membership lists are not
 // validated against real records, so a junk empty entry must not stand in for
 // a principal that has no organization user record yet.
+//
+// Subjects are matched issuer-qualified, so this is the predicate for writes
+// that carry a client-authored subject record: a record at a new issuer is a
+// new stored fact and callers may treat it as an addition.  A gate asking
+// whether the group's roles are already conferred must use HasMemberByID
+// instead.
 func (s *GroupSpec) HasMember(organizationUserID string, subject GroupSubject) bool {
 	if organizationUserID != "" && slices.Contains(s.UserIDs, organizationUserID) {
 		return true
 	}
 
 	return s.HasSubject(subject)
+}
+
+// HasMemberByID reports whether the group already confers its roles on the
+// principal, matching subjects by ID alone.  This mirrors how RBAC actually
+// resolves membership: its subject matching deliberately ignores the recorded
+// issuer (see groupSubjectFilter in pkg/rbac), because subject records
+// written before issuers were recorded carry an empty one and must still
+// resolve.  A grant gate has to match the same way — a membership stored as a
+// legacy record already confers the roles, so a write that re-states it
+// confers nothing and must not read as an addition and be refused.  If RBAC
+// matching ever becomes issuer-qualified, this must move with it.
+//
+// An empty subject ID matches nothing, for the same reason as the empty
+// organization user ID above.
+func (s *GroupSpec) HasMemberByID(organizationUserID, subjectID string) bool {
+	if organizationUserID != "" && slices.Contains(s.UserIDs, organizationUserID) {
+		return true
+	}
+
+	if subjectID == "" {
+		return false
+	}
+
+	return slices.ContainsFunc(s.Subjects, func(subject GroupSubject) bool {
+		return subject.ID == subjectID
+	})
 }
--- a/pkg/handler/README.md
+++ b/pkg/handler/README.md
@@ -163,6 +163,12 @@ Across the package, the common limitations are:
 
 This is a systemic property of the storage model, not just an isolated flaw in one client.
 
+Authorization is deliberately kept out of that gap on the paths that write group membership.
+`users`, `serviceaccounts` and `groups` settle every grant a request makes before the request
+writes anything, so a refusal applies none of it. What can still leave a partial write is
+infrastructure failure partway through, which is the limitation above and not an authorization
+one. See [`users`](./users/README.md) and [`serviceaccounts`](./serviceaccounts/README.md).
+
 ## Package Map
 
 - [`organizations`](./organizations/README.md): tenancy-root visibility and current `v1`
--- a/pkg/handler/common/membership.go
+++ b/pkg/handler/common/membership.go
@@ -19,6 +19,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	servererrors "github.com/unikorn-cloud/core/pkg/server/errors"
 	unikornv1 "github.com/unikorn-cloud/identity/pkg/apis/unikorn/v1alpha1"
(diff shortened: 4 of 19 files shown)


spjmurray:
The gate does what it says, on both paths. Re-ran the `main`-branch hole against this branch — `PUT /users` with `spec.groupIDs` naming a group carrying an ungrantable role:

```
update err = members cannot be added to the group: role "radar" (radar-id) is not grantable by the caller
create err = members cannot be added to the group: role "radar" (radar-id) is not grantable by the caller
group after: userIDs=[] subjects=[]
```

Same on the service-account path, and the pre-pass ordering holds: nothing is written before the refusal, so a rejected create strands neither a global user nor a service account with a token already issued.

Also swept for a route this misses. Only three files write group membership — `pkg/handler/groups`, `pkg/handler/users`, `pkg/handler/serviceaccounts` — and all three are gated now. `pkg/handler/projects` stores `GroupIDs` on the Project and adds finalizers; it never touches a group's member lists. The rewritten trust assumption in `allowGrantProjectScope` enumerating all three routes is the right shape, and correcting the stale `groupSubjectFilter` comment is overdue.

One thing to settle before this goes in.

## The write and read paths still use the old predicate

#529 established that a comparison deciding whether the group already confers its roles has to match how `pkg/rbac` resolves membership — subject ID alone, issuer ignored. That landed on the gate. `addToGroup`, `removeFromGroup` and `convert` were left on issuer-qualified and `UserIDs`-only comparisons respectively.

With a member stored as a legacy empty-issuer subject — the shape `HasMemberByID` exists for — a full add/remove cycle through `PUT /users`:

```
after add:    userIDs=[orguser-alice] subjects=[{alice@example.com  alice@example.com} {alice@example.com https://identity.unikorn-cloud.org alice@example.com}]
after remove: userIDs=[]              subjects=[{alice@example.com  alice@example.com}]
```

The add appends a second record for the same principal, because `HasSubject` is issuer-qualified and does not see the legacy one. The remove then deletes only the record it wrote. Alice is still a member as far as RBAC is concerned, so she keeps `radar:things`, and the call returned 200 with `groupIDs: []` in the body. **That is a revocation that reports success and does not revoke.**

Third piece, same root: `convert` builds the reported `GroupIDs` from `Spec.UserIDs` alone, so a subject-only membership is invisible over the API:

```
List reports alice groupIDs=[] (she is an RBAC member of group-alpha)
```

All three pre-date this PR — `main` has the same `removeFromGroup` comparison and the same `convert` loop, so none of it is a regression here. Raising it here because this is the change that made "match how RBAC resolves membership" the rule, and it applied it to one of the four places that needs it. Leaving the other three is how the next person concludes the rule is only about gates.

Simplest fix, and the one I would take: use the ID-only predicate on the write path too. `removeFromGroup` deletes any subject record with that ID, `addToGroup` treats any record with that ID as already present. Removal then actually removes, the duplicate never appears, and there is one notion of "is this principal a member" in the package instead of two. `convert` becomes `group.Spec.HasMemberByID(in.Name, user.Spec.Subject)`.

The cost is that a legacy record then never gets upgraded to an issuer-qualified one, since nothing rewrites it. If you would rather migrate the data, have `addToGroup` replace a matching legacy record rather than skip it — but that is a bigger change and I would not hold this PR for it.

Either way, or scoped out to its own issue with the revocation failure written down — happy with any of those. Not approving until the revocation part has an answer, since that one is silent.

spjmurray (approved):
All three parts fixed, and taking the simple option — one ID-only predicate for gate, write and read — was the right call. Deleting `Matches` and `HasSubject` rather than keeping a second notion of "same member" alongside `HasMemberByID` is what makes it stay fixed. Keeping `IdentityKey` for the groups path's request-subject dedup is a defensible split, and the commit says why.

Re-ran the probes on `71a24c7e`.

Revocation now revokes:

```
after remove: subjects=[] userIDs=[] response=[]
```

Subject-only membership is reported: `List reports groupIDs=[group-alpha]`. Read-then-write-back keeps it, and fills in the `UserID` half: `sent=[group-alpha] -> subjects=[{alice@example.com  alice@example.com}] userIDs=[orguser-alice]`. Those two together are what matters — with the old `convert` and the new `removeFromGroup`, a client round-tripping what it read would have silently revoked a subject-only membership. Fixing both in one commit avoids that window.

The gate itself is still closed, on both entry points, with nothing left behind by a refusal:

```
update err = members cannot be added to the group: role "radar" (radar-id) is not grantable by the caller
create err = members cannot be added to the group: role "radar" (radar-id) is not grantable by the caller
group after: subjects=[] userIDs=[]
global users after refused create: 1     (bob only — no record for the refused subject)
```

Service accounts the same, and the pre-pass ordering holds there too.

Swept once more for a route this misses. Three files write group membership — `groups`, `users`, `serviceaccounts` — and all three are now gated. `projects` stores `GroupIDs` on the Project and manages finalizers; it never touches a member list. `golangci-lint` clean, `make generate` leaves no diff, `go test ./pkg/...` green across 22 packages.

Approving. Two notes for the record, neither blocking.

## Consequence of the ID-only removal

Matching by ID alone means a removal now deletes every subject record with that ID, including one stored at a different issuer:

```
group before: subjects=[{alice@example.com https://partner.example.com}, {bob@example.com https://partner.example.com}]  userIDs=[orguser-alice]
remove LOCAL alice
group after:  subjects=[{bob@example.com https://partner.example.com}]  userIDs=[]
```

The partner-issuer record for the same subject string goes with it. That is consistent with `pkg/rbac`, which already treats both records as the same member, so it is not a behaviour change in who holds what — and it needs `identity:users` write, which can strip memberships anyway, so there is no authority to gain from it. Worth knowing it happens, because the groups path still dedups request subjects issuer-qualified, so that path can create two records the users path will remove together.

## The thing underneath

`groupSubjectFilter` ignoring the issuer is now pinned in four places, with a comment in each saying they must move together if it changes. That is the right hedge, but the reason it needs hedging is that a subject string is being treated as a globally unique principal name across every federated issuer. Two principals at different issuers with the same subject string are one member as far as RBAC is concerned. Whether that is acceptable depends entirely on whether every trusted issuer's subject namespace is administratively controlled, which is the same assumption `globalGroupRoleBindings` already documents at length. Not this PR's job, and this PR does not widen it — but it is the last API-reachable way I can find to end up holding a role nobody granted you, so it deserves its own issue rather than four comments.

PR #533, opened 2026-08-07
feat(rbac): generic global role bindings replace the platform-admin fast-path (ID-398)

andreyyantsen-nscale:
Replaces the platform-administrator fast-path in `processUserAccountACL` with one generic mechanism — a **global role binding** mapping an `(issuer, subject | "*")` pair to a set of role IDs — configured via a repeated `--global-role-binding` flag and a `globalRoleBindings` chart value.

## Motivation

Global authority is a hardcoded special case. `processUserAccountACL` short-circuits before role resolution: if `(src_iss, subject)` matches a `--platform-administrator-subjects` entry, the caller gets the global permissions of `--platform-administrator-role-ids` and RBAC returns early (`pkg/rbac/rbac.go:722` on `main`).

Two consequences:

- Each new kind of global principal — ID-399's platform-wide staff reader is the next one — needs another bespoke fast-path and flag, duplicating the issuer URL into deployment config, where the verbatim, trailing-slash-sensitive `iss` match is easy to get wrong.
- The decision lives outside RBAC, so it cannot move cleanly to the policy engine (Cerbos Phase 1).

This replaces the fast-path with one mechanism: `(issuer, subject | "*") → [role IDs]`. Platform administrators become one binding of that kind, ID-399's staff reader another, and neither needs Go changes.

## Semantics

- **Replace, not merge.** A matched principal skips membership resolution for that session, so org-scoped writes are absent while authenticated via a bound issuer. This is session-level privilege separation, and it is the behaviour the direct integration assertion pins.
- **Wildcards are clamped to read** at accumulation time, and never match the `uni` sentinel or an unset issuer — so impersonated principals and pre-`src_iss` passports fail closed. The clamp bounds *verbs, not payload sensitivity*: read endpoints can still return credential material, so any role referenced by a wildcard binding needs a read-surface audit first.
- **Issuer-wide bindings are only appropriate** for issuers whose entire user population is itself an authorization decision.

## Compatibility

Compatible. The legacy `--platform-administrator-*` flags are translated verbatim into exact bindings and keep working unchanged — a legacy subject literally `*` stays an exact match and never widens to wildcard semantics. `--platform-administrator-role-ids` keeps its meaning as the role set granted to those subjects, and no chart value changed meaning.

One symbol is removed: `rbac.NewSuperContext`, which lost its only caller when onboarding was deleted (#491) and has none in any nscaledev repo — verified by org-wide code search. It was the last non-binding reader of `PlatformAdministratorRoleIDs`, so dropping it puts the legacy flags on a real removal path. Every other exported symbol is unchanged.

A unit test pins that the legacy flags and equivalent bindings produce identical ACLs, and the integration suite proves it against a live cluster with two bound identities, one per mechanism.

## Two hardening fixes riding along

- **`src_iss` in the ACL cache key** (ID-367 finding 6). Bindings make the authenticated issuer authorization-relevant, so the old key let the same subject value from two issuers share a cached ACL. Every user-influenced segment is length-prefixed rather than plainly joined, because subjects routinely contain the `|` delimiter (Auth0 emits `auth0|<id>`) and an unescaped join lets one identity forge another's key.
- **Inactive users rejected on the bearer path**, regardless of `allowExternalIdentity`. That flag is meant to admit subjects UNI has never seen, not to re-admit one it has explicitly deactivated. `ErrUserInactive` wraps `ErrResourceReference`, so every existing `errors.Is` caller behaves as before.

## Reviewer notes

- **Misconfiguration fails the Helm render, not the pod.** Subjects are trimmed once before every guard reads them, and `::` — the flag separator, which would silently shift the right-anchored parse — is rejected on both issuer and subject. The chart is the only layer that can catch the separator case: once the flag string is built the mis-parse is undetectable server-side. `hack/check_chart_render.sh` (wired into `make lint`) asserts each negative case fails for *its own* reason, so one broken guard can't hide behind another's error.
- **Role-set drift is the one class the render cannot cover, and it fails closed.** A role name absent from the chart's catalogue produces a perfectly valid flag that then 500s every user of that issuer at request time; a `Role` CRD deleted *after* a successful render does the same, because roles are live CRDs read per request. Erroring beats silently resolving a smaller ACL on an authorization path, so this is deliberate — but it is an availability cliff worth knowing about before referencing a role from a wildcard binding.
- **Known gaps, deliberately not closed here.** Both concern the inactive-user rejection, and both are documented in `docs/multi-issuer-token-contract.md`:
  - Organization-level suspension takes a different path that resolves to an empty `orgIds` list with no error, so a user suspended in every org is still admitted. Flagged for ID-399, which would build wildcard bindings directly on top of it.
  - The rejection can only fire on a record the lookup finds, and `GetUser` compares `spec.subject` verbatim while the bearer path lower-cases the email claim first. A `User` stored with a mixed-case subject is never matched and is admitted as never-onboarded, while bindings — which match case-insensitively — still grant it authority. The matcher's `EqualFold` is verbatim the legacy fast-path's, so this branch makes a pre-existing gap visible rather than introducing one. Closing it means normalising `GetUser` or `spec.subject` at creation: a change to global identity resolution with a duplicate-record migration attached, tracked separately.
- Commits are self-contained and file-disjoint; each builds and vets independently (`git rebase --exec`), so the branch is bisectable.

## Verification

- `make touch license validate lint generate` with a clean tree, plus `make test-unit`.
- `go vet -tags integration ./test/api/...`.
- Full `make integration-test` against fresh KinD: **192/192 specs, 0 failed, 0 skipped**, including four new specs covering both binding mechanisms and the global-only ACL that detects an accidentally-additive rewrite. That run predates the review fixes above; CI covers the current head.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,6 +200,7 @@ lint: $(GENDIR)
 	@go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(LINT_VERSION)
 	$(GOBIN)/golangci-lint run ./...
 	helm lint --strict charts/identity
+	./hack/check_chart_render.sh
 
 # Validate the server OpenAPI schema is legit.
 .PHONY: validate
--- a/charts/identity/templates/identity/deployment.yaml
+++ b/charts/identity/templates/identity/deployment.yaml
@@ -43,6 +43,42 @@ spec:
           {{- $systemAccounts = append $systemAccounts (printf "%s=%s" $k (include "resource.id" $v)) }}
         {{- end }}
         - --system-account-roles-ids={{ join "," $systemAccounts }}
+        {{- range $i, $b := .Values.globalRoleBindings }}
+          {{- if not (trim (default "" $b.issuer)) }}
+            {{- fail (printf "globalRoleBindings[%d]: issuer is required" $i) }}
+          {{- end }}
+          {{- if and (hasKey $b "subject") (hasKey $b "subjects") }}
+            {{- fail (printf "globalRoleBindings[%d]: subject and subjects are mutually exclusive" $i) }}
+          {{- end }}
+          {{- $subjects := list }}
+          {{- if hasKey $b "subject" }}
+            {{- $subjects = list $b.subject }}
+          {{- else }}
+            {{- $subjects = default (list) $b.subjects }}
+          {{- end }}
+          {{- if not $subjects }}
+            {{- fail (printf "globalRoleBindings[%d]: exactly one of subject or non-empty subjects required" $i) }}
+          {{- end }}
+          {{- range $si, $subject := $subjects }}
+            {{- if not (trim (default "" $subject)) }}
+              {{- fail (printf "globalRoleBindings[%d].subjects[%d]: empty or whitespace-only subject" $i $si) }}
+            {{- end }}
+          {{- end }}
+          {{- if not $b.roles }}
+            {{- fail (printf "globalRoleBindings[%d]: roles must be non-empty" $i) }}
+          {{- end }}
+          {{- $allRoles := merge (dict) (default (dict) $.Values.roles) (default (dict) $.Values.additionalRoles) }}
+          {{- $roles := list }}
+          {{- range $b.roles }}
+            {{- if not (hasKey $allRoles .) }}
+              {{- fail (printf "globalRoleBindings[%d]: unknown role %q" $i .) }}
+            {{- end }}
+            {{- $roles = append $roles (include "resource.id" .) }}
+          {{- end }}
+          {{- range $subject := $subjects }}
+        - {{ printf "--global-role-binding=%s::%s::%s" $b.issuer $subject (join "," $roles) | quote }}
+          {{- end }}
+        {{- end }}
         {{- range $i, $arg := .Values.server.extraFlags }}
         - {{ $arg }}
         {{- end }}
--- a/charts/identity/templates/identity/deployment.yaml
+++ b/charts/identity/templates/identity/deployment.yaml
@@ -37,6 +37,21 @@ spec:
           {{- $adminRoles = append $adminRoles (include "resource.id" $name) }}
         {{- end }}
         - --platform-administrator-role-ids={{ join "," $adminRoles }}
+        {{- range $i, $s := .Values.platformAdministrators.subjects }}
+          {{- /*
+          The server matches subjects case-sensitively; only the "::"-separated
+          subject part is checked, mirroring how the flag parser splits issuer
+          from subject.
+          */}}
+          {{- $subj := $s }}
+          {{- $cut := splitn "::" 2 $s }}
+          {{- if $cut._1 }}
+            {{- $subj = $cut._1 }}
+          {{- end }}
+          {{- if and (ne $subj "*") (regexMatch "[A-Z]" $subj) }}
+            {{- fail (printf "platformAdministrators.subjects[%d]: subject must be its canonical lower-case form" $i) }}
+          {{- end }}
+        {{- end }}
         - --platform-administrator-subjects={{ join "," .Values.platformAdministrators.subjects }}
         {{- $systemAccounts := list }}
         {{- range $k, $v := .Values.systemAccounts }}
@@ -82,6 +97,9 @@ spec:
             {{- end }}
             {{- if contains "::" $trimmed }}
               {{- fail (printf "globalRoleBindings[%d].subjects[%d]: subject must not contain \"::\"" $i $si) }}
+            {{- end }}
+            {{- if and (ne $trimmed "*") (regexMatch "[A-Z]" $trimmed) }}
+              {{- fail (printf "globalRoleBindings[%d].subjects[%d]: subject must be its canonical lower-case form" $i $si) }}
             {{- end }}
             {{- if and (eq $b.issuer "uni") (eq $trimmed "*") }}
               {{- fail (printf "globalRoleBindings[%d]: wildcard subject not allowed on the UNI sentinel issuer" $i) }}
--- a/charts/identity/templates/identity/deployment.yaml
+++ b/charts/identity/templates/identity/deployment.yaml
@@ -51,11 +51,9 @@ spec:
             {{- fail (printf "globalRoleBindings[%d]: issuer must not contain whitespace" $i) }}
           {{- end }}
           {{- /*
-          The server parses issuer::subject::roles right-anchored on the last
-          two separators, so a "::" anywhere in the issuer or subject silently
-          shifts the boundary. Only the chart still knows the intended split —
-          once the flag string is built the mis-parse is undetectable — so both
-          sides are rejected here rather than in pkg/rbac.
+          Right-anchored parsing means a "::" in either field silently shifts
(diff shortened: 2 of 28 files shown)


spjmurray:
## Review notes — architecture, not blocking

Read against the platform security posture and RBAC direction docs. Verdict: this merges. Two things for now, three for later.

I verified the compatibility claim rather than taking it on trust: `effectiveGlobalRoleBindings` translates each `--platform-administrator-subjects` entry to `GlobalRoleBinding{Issuer, Subject, RoleIDs: PlatformAdministratorRoleIDs}` with `Wildcard` false, and `resolveGlobalRoleBindings` then applies the same issuer equality and the same `EqualFold(TrimSpace(...))` predicate as `rbac.go:723` on `main`, routing to the unclamped accumulator. Same match, same roles, same early return. No existing platform administrator loses anything, and the replace semantics is pre-existing behaviour this PR names rather than introduces (the `return acl, nil` is a context line in the diff).

`src_iss` in the ACL cache key is the right call and the `auth0|<id>` forging argument for length-prefixing is correct.

### Now

**1. `EqualFold` → `==` in the binding matcher, plus a render guard rejecting non-lower-case subjects.**

Four sites compare the same subject value:

| Site | Comparison |
|---|---|
| `pkg/oauth2/auth0/validator.go:195` | `ToLower` + `TrimSpace` |
| `pkg/userdb/users.go:58` (`GetUser`) | `==` |
| `pkg/rbac/rbac.go` (`groupSubjectFilter`) | `==` |
| `resolveGlobalRoleBindings` | `EqualFold` + `TrimSpace` |

The binding is the only site more permissive than the rest, and it is the one granting global CRUD. So a `User` stored with a mixed-case subject already fails closed everywhere else — no `GetUser` match, no group match, empty `orgIds`, no authority — and fails *open* through the binding, taking the new inactive-user rejection with it (`GetActiveUser` calls `GetUser` first, so the state check never runs).

This suggests the fix is the opposite direction to the one recorded in `docs/multi-issuer-token-contract.md`. Normalising `GetUser` or `spec.subject` at creation needs the duplicate-record migration noted there, and folding the lookup would make `GetUser`'s `List` + `slices.IndexFunc` first-match resolve non-deterministically between two case-variant records. Tightening the grant instead is one operator, no migration, and inverts the failure direction: a mixed-case values entry stops matching, that admin loses access loudly, someone fixes the chart. `EqualFold` is unreachable on the Auth0 path anyway since the validator already lower-cased — its only live effect is absorbing a typo in a chart value, which is a poor trade on an authorization grant.

Agreed this gap predates the branch. What changes is that this PR adds the inactive-user rejection and presents it as a control, and a labelled-best-effort security control tends to get relied on — and the population able to be deactivated-yet-authorised goes from a handful of named admins to an entire issuer's roster once a wildcard binding lands.

*Precondition:* query for any non-lower-case `platformAdministrators.subjects` entry or `User.spec.subject` in a live environment. Those work today *because* of `EqualFold` and would break. If they are all lower-case — likely, given the validator — this is free.

**2. One line in `values.yaml`.** The `subjects: [alice@example.com, bob@example.com]` / `roles: [platform-administrator]` example reads as the live way to grant global CRUD. If naming subjects explicitly is legacy and development-only, say so in the comment. Consistent with this PR's own principle that values.yaml is where an operator learns what is and is not allowed.

### Later

**3. Move the population grant onto `OAuth2Provider.Spec.BearerTrust`.**

The binding's issuer component is a verbatim URL hand-typed into deployment config a second time — which is what this PR's own description identifies as easy to get wrong. That record already exists as the per-issuer trust configuration. Adding a functional population field to it, and binding roles to the population rather than the `iss` string, would remove the reason most of the guard surface exists: `errIssuerNotURL`, `errIssuerCommaOrSpace`, `errIssuerPathSep`, the IPv6-host carve-out, the right-anchored `::` parse, the trailing-slash hazard, and `errSentinelWildcard` (the sentinel simply would not be classified). `ErrUntrustedBindingIssuer` and `computeTrustedNonUNIIssuers` go too — an untrusted issuer becomes unreferenceable by construction, which beats advisory-log-and-continue.

The larger win is that it moves the grant from a Helm flag to a CRD: runtime-mutable, per-issuer revocation, no redeploy. That addresses most of the revocation story in `docs/platform-reader.md` — the incident lever stops being role deletion, which that doc rightly describes as 500-ing every user of the issuer at once.

It also makes a provider migration a re-mapping of one resolution record rather than a re-keying of every grant, and it means one population per issuer, so a claim-gated split inside a shared tenant is structurally impossible rather than merely discouraged.

Depends on tightening `administrator`'s org-scoped `identity:oauth2providers` scope first — see my note on #536. Trust and grant on one object is only safe if that object is platform-writable.

**4. Entity-ID keying, platform-wide.** Email-as-subject is the real exposure and it is universal — a reassigned address inherits group memberships and the `User` record, not just a binding. Not this PR's problem, but bindings are the case worth migrating first: a stale group membership is org-scoped and visible through the API, whereas a stale exact binding is unclamped global CRUD in a values file.

**5. Worth recording:** there is no entity-keyed path to global authority anywhere in the codebase. `acl.Global` is produced only by global role bindings and `processSystemAccountACL`'s X.509 CN; group memberships never yield it. So "make platform administrators a group instead" is not currently available, which is the real justification for exact bindings and worth stating as the gap it is.

### Not changing

The two-layer read clamp is not redundant and should stay: the render guard cannot see a `Role` CRD that gains write scopes afterwards, and the runtime clamp cannot see operator intent at config time.

One deployment note for the rollout, since it is not a break but will bite: the flag is only emitted when `globalRoleBindings` is non-empty and older binaries reject unknown flags — binary first, then enable bindings, reverse on rollback. Already covered in the values.yaml comment; flagging it for whoever runs the upgrade.

andreyyantsen-nscale:
@spjmurray Thanks — and for verifying the compatibility claim and the src_iss reasoning directly. Both "Now" items are done in this branch; the three "Later" items are filed as follow-ups.

Now 1 — EqualFold → == + render guard. Done, in the direction you argued. The matcher is now == on the trimmed subject, and the chart fails to render a non-lower-case subject on both globalRoleBindings and platformAdministrators.subjects, with render-check coverage — so a mixed-case entry breaks loudly at deploy instead of matching silently. I didn't fold GetUser/spec.subject (needs the migration you noted, and makes first-match non-deterministic); tightening the grant is the better trade. Docs updated to the narrowed gap. Still worth running your precondition before rollout: no non-lower-case platformAdministrators.subjects / User.spec.subject live — the guard only covers new chart values.

Now 2 — the values.yaml example. Clarified rather than marked legacy: per your item 5, per-subject bindings are the current and only route to global write, so the comment now says exactly that.

Later 3 — population grant on BearerTrust. Agreed, filed. Captured the full argument: binding to the population dissolves most of the issuer guard surface, moves the grant to a runtime-mutable CRD with per-issuer revocation, and makes migration a one-record remap. Noted the dependency on tightening administrator's identity:oauth2providers scope first.

Later 4 + 5 — entity-ID keying + the gap. Filed together, with your point that bindings are the case worth migrating first, and item 5's finding that no entity-keyed path to global authority exists today (so "make admins a group" isn't available — the real justification for exact bindings).

spjmurray (approved):

PR #536, opened 2026-08-08
feat(chart): add a protected platform-reader role (ID-399)

andreyyantsen-nscale:
Stacked on #533 — do not merge before it.

Adds `platform-reader` to the chart role catalogue: a protected, global, read-only role for support tooling, and the first real consumer of #533's wildcard bindings. Zero production Go — policy data plus guardrails.

## Motivation

Support tooling needs read-only visibility across every organisation. The only cross-organisation role today is `platform-administrator` (full CRUD everywhere), far too dangerous to hand out for that purpose.

Read-only-ness alone is not safety: #533's clamp bounds *verbs, not payload sensitivity*, so a global `read` can still return credentials. Hence a systematic read-surface audit across uni-identity, uni-region, uni-kubernetes, uni-compute and uni-storage — and a contract test that pins the role **definition**, not observed ACL output, which the clamp would mask.

## Definition

The read-projection of `platform-administrator`'s global block: 44 read-bearing scopes minus 6 credential-bearing exclusions and 7 vacuous omissions, leaving **31 scopes, all `[read]`**, `protected: true`, global-only.

Excluded: `region:identities`, `region:servers`, `kubernetes:clusters`, `kubernetes:virtualclusters`, `compute:instances`, `storage:objectstorageendpoints/accesskeys`. Omissions are fail-closed so a future API forces fresh classification.

The contract test ratchets: every read-bearing admin scope must land in exactly one of the three sets with a rationale, and a stale entry fails the build.

## Compatibility

Compatible, purely additive. No existing role, flag, API surface, or production Go changed. The role ships **unbound** — the issuer-wide binding is deployment config in the environment repo, so merging grants no privilege.

## Reviewer notes

- **`identity:users` global read is deliberate**: cross-organisation user lookup is the point of support tooling.
- **Revocation is IdP-only.** Organisation suspension is a silent no-op for a bound subject, and with `allowExternalIdentity: true` there is usually no global `User` record to suspend. Documented, along with #533's case-sensitivity gap.
- **The ratchet covers the identity catalogue only** — a downstream service adding a credential-returning GET under an included scope is invisible to it. One of six follow-ups in `docs/platform-reader.md#follow-ups`.
- **`protected: true` does less than the design spec claimed**: there is no role delete endpoint, so it blocks nothing there. It filters the role from the user-facing listing and blocks granting it through a group. The doc states the real mechanisms.
- The canary binds via an **exact `uni` binding**, bypassing the wildcard clamp, so a write verb in the role fails the suite instead of being masked.
- Commits are file-disjoint and each builds independently; the branch is bisectable.

## Verification

- `make touch license validate lint generate` clean, `make test-unit`, `go vet -tags integration ./test/api/suites/`.
- `hack/check_chart_render.sh`, now with the first *non-vacuous* positive case for the wildcard guard (the existing `reader` case has no global block).
- Full `make integration-test` on fresh KinD: **197/197 specs, 0 failed, 0 skipped**, including five new platform-reader specs.
- Every guardrail mutation-tested: a write verb fails the contract test and the render guard; un-redacting `clientSecret`/`accessToken` fails the redaction tests; a removed scope fails the ratchet.

### DIFF
--- a/charts/identity/values.yaml
+++ b/charts/identity/values.yaml
@@ -156,7 +156,8 @@ platformAdministrators:
 # per-subject bindings here are the only route to global write authority.
 # Full semantics: pkg/rbac/README.md#global-role-bindings.
 # Helm rejects invalid separators and wildcard bindings on roles that grant any
-# non-read global operation — no role shipped below qualifies for wildcard use.
+# non-read global operation — platform-reader is the only role shipped below
+# with global scopes that qualifies for wildcard use.
 # Rollout ordering: the flag is emitted only when this list is non-empty, and
 # older server binaries do not recognise it — deploy a binary that understands
 # --global-role-binding before enabling bindings; on rollback, clear this
@@ -164,7 +165,7 @@ platformAdministrators:
 globalRoleBindings: []
 #  - issuer: https://staff.example.auth0.com/
 #    subject: "*"
-#    roles: [platform-reader] # example only: role must already exist in roles/additionalRoles
+#    roles: [platform-reader]
 #  - issuer: uni
 #    subjects: [alice@example.com, bob@example.com]
 #    roles: [platform-administrator]
@@ -232,6 +233,47 @@ roles:
         storage:objectstorageclasses: [create,read,update,delete]
         storage:objectstorageendpoints: [create,read,update,delete]
         storage:objectstorageendpoints/accesskeys: [create,read,update,delete]
+  # A platform-reader has global read-only visibility for support tooling.
+  # Definition = read-projection of platform-administrator's global scopes
+  # minus credential-bearing and unserved scopes. The projection, the
+  # exclusion list, and read-only-ness are pinned by the contract test in
+  # pkg/rbac (see docs/platform-reader.md for the audit record and runbook).
+  platform-reader:
+    description: Platform reader
+    protected: true
+    scopes:
+      global:
+        identity:organizations: [read]
+        identity:oauth2providers: [read]
+        identity:roles: [read]
+        identity:serviceaccounts: [read]
+        identity:users: [read]
+        identity:groups: [read]
+        identity:projects: [read]
+        identity:quotas: [read]
+        identity:allocations: [read]
+        region:regions: [read]
+        region:flavors: [read]
+        region:images: [read]
+        region:externalnetworks: [read]
+        region:networks: [read]
+        region:networks:v2: [read]
+        region:loadbalancers:v2: [read]
+        region:securitygroups: [read]
+        region:securitygroups:v2: [read]
+        region:volumeclasses:v2: [read]
+        region:filestorage:v2: [read]
+        region:filestorageclass:v2: [read]
+        region:sshcertificateauthorities:v2: [read]
+        kubernetes:regions: [read]
+        kubernetes:flavors: [read]
+        kubernetes:images: [read]
+        kubernetes:clustermanagers: [read]
+        compute:regions: [read]
+        compute:flavors: [read]
+        compute:images: [read]
+        storage:objectstorageclasses: [read]
+        storage:objectstorageendpoints: [read]
   region-service:
     decription: Region service
     protected: true
--- a/docs/platform-reader.md
+++ b/docs/platform-reader.md
@@ -1,0 +1,209 @@
+# The `platform-reader` Role
+
+`platform-reader` is a **protected**, global-only, read-only role in the identity role catalogue.
+It exists so staff support tooling can gain cross-organization read visibility without the
+blast radius of `platform-administrator` (full CRUD everywhere). It is the first real wildcard
+consumer of the generic global-role-binding mechanism landed by ID-398 (PR #533): every
+wildcard-specific runtime behaviour — the render-time guard, the runtime read clamp, the
+issuer-wide blast radius — goes live in production for the first time with this role.
+
+The role's definition is the **read-projection of `platform-administrator`'s global scope block**,
+minus an explicit, machine-checked exclusion list and a fail-closed omission list for scopes
+nothing currently serves. That projection, the exclusion/omission classification, and the
+role's read-only-ness are pinned by `pkg/rbac/platform_reader_contract_test.go`: any future PR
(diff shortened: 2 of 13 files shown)


spjmurray:
## Review notes — architecture, not blocking

Read alongside #533 against the platform security posture and RBAC direction docs. This merges. The read-surface audit and the decision to ship the role unbound are both the right calls, and the follow-up list being honest about what the ratchet cannot see is what made the rest of this review possible.

### Now

**1. Drop `administrator`'s org-scoped `identity:oauth2providers` from `[create,read,update,delete]` to `[read]`.**

Currently in `charts/identity/values.yaml`, the ordinary organization administrator role holds full CRUD on oauth2providers at organization scope — the bring-your-own-IdP aspiration, which is no longer a live use case. Two reasons to tighten it here rather than later:

- It is a denial of service today. An organization administrator can delete their own organization's provider and lock every user in that organization out of login.
- It is the escalation that opens the moment the provider record carries any authority-relevant field (see my note on #533, item 3). Create a provider, declare its population, receive global read of every organization — starting from an ordinary customer role. Trust and grant on one object only works if the object is platform-writable.

Read stays safe precisely because this PR pinned secret redaction in `pkg/handler/oauth2providers/client_internal_test.go`, which is also what makes `platform-reader`'s global `identity:oauth2providers: [read]` defensible. `auditor` already has organization-scoped read, so this just makes the two roles agree; `platform-administrator` keeps global CRUD, and global-only means it is not reachable through organization membership.

*Precondition:* confirm no `OAuth2Provider` objects currently live in organization namespaces rather than the platform namespace. Existing objects keep working either way, but anyone self-serving federation would lose the ability to edit.

**2. Name role-emptying as the deliberate incident lever in the runbook.** `docs/platform-reader.md` documents role deletion only under "Blast radius" as a hazard. It is also the only fast, self-owned revocation lever the platform has — the Revocation section otherwise leaves IdP offboarding as the sole option. Worth an explicit step: what to run, that it fails closed as a 500 rather than a 403, and that it lands within `acl-cache-timeout` (default 1 minute).

### Later

**3. Per-scope sensitivity marker, owned by the defining service; derive `platform-reader` from it.**

The role is *defined* as a projection — read-bearing global scopes of `platform-administrator`, minus credential-bearing, minus unserved — but *stored* as 31 hand-listed scopes with a contract test pinning the relationship. If the relationship is the truth, expressing the relationship and letting the 31 fall out is the stronger shape. The blocker is that "read but not credential-bearing" is a fact that exists nowhere, which is exactly why the list had to be hand-built.

Putting that fact on the scope, declared by whoever owns it, would: remove the hand-list; remove the vacuous-omission set entirely, since absent from the catalogue means absent from the projection; and change the ratchet's question from "did anyone remember to classify this" to "is this marked correctly", which the owning team can answer and this repo cannot.

That also makes the first follow-up in `docs/platform-reader.md` closable rather than permanent — a downstream service adding a credential-returning GET under an included scope marks it at the point of adding it.

The argument for doing it: as it stands, every new API in region, compute, kubernetes or storage owes this repo a written classification or the build breaks, and that lands on whoever is on call in identity — the person least placed to answer whether an endpoint in someone else's service returns credentials. The ratchet converts silent drift into a loud failure, which is better, but in the wrong repo.

**4. The four downstream splits are what make the role useful, not tidying.** With `kubernetes:clusters`, `kubernetes:virtualclusters`, `compute:instances` and `region:servers` all excluded, a holder can see organizations, users, groups, projects, quotas, allocations and the infrastructure catalogue — and no customer workload at all. That answers "who is this customer and what have they bought" but not "what is wrong with their cluster", which is what support tends to get paged about. Worth being explicit with whoever asked for this that day one is an organization-structure viewer, and that the workload visibility depends on work in three other repositories.

**5. Raise the region v2 power-actions defect as its own issue in its own repository.** The audit found that v2 start/stop/reboot are gated by the read check rather than an update check as v1 requires. That is a live authorization defect in another service, and it should not live as a bullet in an identity PR's follow-up list where nobody owning that code will see it.

andreyyantsen-nscale:
@spjmurray Thanks — all five taken. Status:

1 — done in #538, stacked on this PR: administrator drops to `[read]` on org-scoped `identity:oauth2providers`, pinned by a mutation-tested grant guard, with two integration specs asserting the denial. Precondition checked: no OAuth2Provider objects exist in organization namespaces in staging or production. Since self-serve federation is no longer a live use case, the follow-up is removing the org-scoped provider API family entirely rather than leaving a read-only stub.

2 — done in 4bb9ae63: the runbook now names role-emptying as the deliberate incident lever, with the command to run, the distinction that emptying scopes denies as clean 403s while deleting the CRD fails closed as 500s, and the latency bound from acl-cache-timeout.

3 — extracted to a ticket, with one refinement from writing it up: the classification does exist as data today — the contract test's exclusion/omission maps — so the real defect is authorship location, not absence. And fully closing the "new credential GET under an included scope" class needs the marker at per-operation granularity, not per-scope; a scope-level marker moves the write-up to the right repo but doesn't mechanically force the re-mark. The ticket carries both constraints plus candidate homes (OpenAPI annotations are the only candidate with natural per-operation granularity).

4 — tracked, with communication as the explicit first deliverable: telling the requesters that day one is an organization-structure viewer (including the topology surface it does grant), and that workload visibility depends on the splits in three other repositories. Each split is also flagged as a two-repo change with grant-before-enforce ordering.

5 — filed as nscaledev/uni-region#648, and verifying it end-to-end turned up a correction and a companion. The correction: no shipped role holds region:servers read without update — reader/auditor have no `region:servers` grant at all — so the region defect is a latent regression rather than live exposure; its cost is that read-only `region:servers` can never be safely granted until fixed. The live exposure is one layer up: uni-compute's instance start/stop/reboot are gated only by `compute:instances` read, which project reader and org auditor do hold, and the proxy calls region as the compute service account with no impersonation, so region-side RBAC never constrains the end user. Filed as nscaledev/uni-compute#321; the two are cross-linked and independently fixable (both fixes verified against the system-account grants — nothing breaks).

spjmurray (approved):

PR #538, opened 2026-08-10
feat(chart): reduce administrator to read on identity:oauth2providers

andreyyantsen-nscale:
Stacked on #536 — do not merge before it. Raised in [review on #536](https://github.com/nscaledev/uni-identity/pull/536#issuecomment-5241044219).

Drops the organization `administrator` from `[create,read,update,delete]` to `[read]` on `identity:oauth2providers`. Policy data plus tests; zero production Go.

## Why

An organization administrator can delete their own provider and lock every user in that organization out of login. And once the provider record carries any authority-relevant field, creating one becomes an escalation path out of an ordinary customer role.

`auditor` already had organization-scoped read, so this just makes the two organization roles agree. `platform-administrator` keeps global CRUD, which is not reachable through organization membership. Nobody uses org-scoped providers in staging or production today.

## Notes for review

- The value is pinned in `pkg/rbac/grant_guard_test.go`, mutation-tested. The grant lattice passes under both `[read]` and CRUD and the render guard only sees global scopes, so nothing fast catches a restored write verb without it.
- Two integration specs assert the denial. They use a synthetic ID — the handler authorizes before lookup — so they need no fixture and never skip.
- The default integration `client` is an `administrator`, so provider writes in the suites moved to `BINDING_ADMIN_AUTH_TOKEN`. Not `PLATFORM_ADMIN_AUTH_TOKEN`: that one rides the transitional `platformAdministrators.subjects` path and would take the specs down with it.
- Revocation lands within `--acl-cache-timeout` (default 1m); no pod restart. Break-glass is Helm, not API — `platform-administrator` is protected.
- Deployments granting provider writes via `additionalRoles` keep resolved ACLs, but group membership edits on such a role start failing (`AllowRole` revalidates on update). Worth a grep before merge.

## Verification

`make touch license validate lint generate` clean, `git status` clean, `make test-unit`, `go vet -tags integration`. Probed live: binding-admin creates (201) and deletes (200) providers.

**The integration suite is red on the base branch too** — untouched `andreyyantsen/id-399` fails 9/197. Two runs here failed 19/199 and 25/199, the second with no failures in any suite this PR touches, and the untouched baseline failed four specs that passed here. Intermittent `401 token validation failed` on user and globally-bound tokens, not expiry. Lives in the global role binding token path; worth chasing separately.

### DIFF
--- a/charts/identity/values.yaml
+++ b/charts/identity/values.yaml
@@ -325,13 +325,15 @@ roles:
         identity:projects/references: [create,delete]
         identity:allocations: [create,read,update,delete]
         region:networks:v2/references: [create,read,update,delete]
-  # An administrator can do anything within an organization.
+  # An administrator can do anything within an organization, except configure
+  # its upstream identity providers: provider writes would let an organization
+  # lock its own users out of login, so they stay with the platform operator.
   administrator:
     description: Organization administrator
     scopes:
       organization:
         identity:organizations: [read,update]
-        identity:oauth2providers: [create,read,update,delete]
+        identity:oauth2providers: [read]
         identity:serviceaccounts: [create,read,update,delete]
         identity:users: [create,read,update,delete]
         identity:roles: [create,read,update,delete]
--- a/hack/ci/fixtures/main.go
+++ b/hack/ci/fixtures/main.go
@@ -558,11 +558,13 @@ func main() {
 
 	// ── Resolve role IDs ─────────────────────────────────────────────────────
 	// platform-administrator is protected and not returned by the API.
-	// We use administrator (org-scoped, full identity CRUD) and user (project-scoped).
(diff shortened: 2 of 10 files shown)


spjmurray:
Just FYI I'm kinda in two minds about this now.  What we are saying with the current model is we control the issuers, some can grant implicit permissions, and thus no-one outside of operations can actually define an issuer.  That's fine for today.

Tomorrow however, a user may NEED to bring their own issuer, or switch it out for a new one, which suggests we should keep the ability to create issuers, but in the new world order we don't have the ability to grant ourselves anything implicitly, the end user holds that right, and only them.

So it's the right thing to to today, but probably the wrong thing tomorrow, thus keeping the mechanism is right, as is limiting the risk while we have implicit grants.  If you catch my drift!  And you Claude when I use you for a review in a minute...

spjmurray (approved):

